Issue #5, opened 2020-04-23
"panicked at 'attempt to divide by zero" when I call functions.


jimmy3dita:
I've sent 500 $NEAR from an account with 500 $NEAR. It succeeded but now I can't anymore interact with the contract.
Deposit TX - https://explorer.betanet.nearprotocol.com/transactions/FjpmJWbdBfWbGctwpJrGQpEViYWGgm7rHKzL64WxXTXp

```
o:nearprotocol pepe$ near call nearkat withdraw '{"amount": "500000000000000000000000000"}' --accountId nuc2.test
Using options: {
  accountId: 'nuc2.test',
  networkId: 'betanet',
  nodeUrl: 'https://rpc.betanet.nearprotocol.com',
  contractName: 'nearkat',
  walletUrl: 'https://wallet.betanet.nearprotocol.com',
  helperUrl: 'https://helper.betanet.nearprotocol.com',
  gas: '100000000000000',
  amount: '0',
  methodName: 'withdraw',
  args: '{"amount": "500000000000000000000000000"}'
}
Scheduling a call: nearkat.withdraw({"amount": "500000000000000000000000000"})
Error:  GuestPanic [Error]: Smart contract panicked: panicked at &#39;attempt to divide by zero&#39;, src&#x2F;lib.rs:103:47
    at Object.parseRpcError (/usr/local/lib/node_modules/near-shell/node_modules/near-api-js/lib/utils/rpc_errors.js:25:19)
    at Account.signAndSendTransaction (/usr/local/lib/node_modules/near-shell/node_modules/near-api-js/lib/account.js:114:36)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async scheduleFunctionCall (/usr/local/lib/node_modules/near-shell/commands/call.js:29:34)
    at async Object.handler (/usr/local/lib/node_modules/near-shell/utils/exit-on-error.js:4:9) {
  type: 'GuestPanic',
  index: 0,
  panic_msg: "panicked at 'attempt to divide by zero', src/lib.rs:103:47"
}
```

evgenykuzyakov:
Seems the issue is here: https://github.com/near/initial-contracts/blob/62bb406180da3f7082e64c0e0eba5448f0bc0b04/staking-pool/src/lib.rs#L103

Issue #3, opened 2020-04-21
`ping` doesn't scale. It has to execute at constant write operations.


evgenykuzyakov:
Current ping version is going to hit the gas limit for `storage_write` limits quite fast. It's likely the limit is for 20-30 users max

Issue #2, opened 2020-04-20
Add a function to return total balance


evgenykuzyakov:
Right now if I want to verify the user doesn't have any balance on the staking pool, I have to call both `get_user_balance` and `get_user_stake`

Issue #1, opened 2020-04-20
Return balance of 0, when user doesn't exist


evgenykuzyakov:
Let's return balance of `0` for both liquid and stake balances if the user doesn't exists.

https://github.com/near/staking-contract/blob/c9f463cfbf5c0d75d9656fde98789d34830ddd2e/src/lib.rs#L169-L178

PR #7, opened 2020-04-28
Update staking pool

evgenykuzyakov:
Rewriting staking pool contract:
- Introduce max number of seats. This is required to prevent internal contract state from growing too big. Accounts with smaller stake are kicked out of the staking epoch and considered inactive after 3 epochs. Inactive accounts are eventually archived, by moving from the internal contract state to the persistent map of archived users.
- The contract never decreases the staked amount for accounts without `unstake` call. This is important for a `lockup` contract to track current unstaked balance. When an account is kicked out, the relative staked amount  for this account doesn't change, but this account becomes inactive.
- Modified the way we calculate rewards to be able properly distribute rewards. For example, if a large account unstakes, the locked amount will eventually drop more than the accumulated rewards. With current logic, such rewards will still be distributed.
- Some small amount of rewards might still be kept by the contract due to rounding errors.

Fixes: #1
Fixes: #2
Fixes: #3
Fixes: #5

It's not properly tested yet (e.g. with real node).


staking-pool/src/lib.rs
```diff
@@ -70,111 +95,354 @@ impl StakingContract {
...
-        assert!(
-            env::state_read::<StakingContract>().is_none(),
-            "Already initialized"
-        );
-        let mut pk = vec![0];
-        pk.extend(bs58::decode(stake_public_key).into_vec().unwrap());
+    pub fn new(
+        owner_id: AccountId,
+        stake_public_key: Base58PublicKey,
+        max_number_of_seats: u32,
```
ilblackdragon, line 101: agree with @jimmy3dita let's not call this seats.
I also not sure we really need this?
May be instead just have field with min stake? This min stake should cover the storage of the delegation?
Also ppl can't really make contract to overfill, because it will just decline if they send a transaction that takes more space then adds funds. So it kinda was working automatically having min stake requirement if there is too much space taken by contract?

I'm not completely sure how you are saving the space here with the active/inactive/archived here?
It seems like archived accounts still need to store accountid & account?

evgenykuzyakov: The issue the iteration within a persistent map. Every lookup/insert requires `storage_read/storage_write`. These operations take too much gas comparing to the on time large `storage_read` because of the `base` fee required to locate the key.

This implementation doesn't issue constant amount of storage read/writes

evgenykuzyakov:
I'll try to modify the staking pool to work without slots to simplify logic.

PR #8, opened 2020-05-01
Token based staking pool

evgenykuzyakov:
Implement internal shares for the staking pool. It simplifies the contract a lot and allows huge amount of accounts to use it.

The reward distribution is not perfect, but over longer time it should work fine.


staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+    pub fn ping(&mut self) {
+        // Checking if we need there are rewards to distribute.
+        let epoch_height = env::epoch_height();
+        if self.last_epoch_height == epoch_height {
+            return;
+        }
+        self.last_epoch_height = epoch_height;
+
+        let total_balance =
+            env::account_locked_balance() + env::account_balance() - env::attached_deposit();
```
ilblackdragon, line 93: why there is attached deposit?
this function is not payable...

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+            .unstaked_available_epoch_height
+            <= env::epoch_height()
+    }
+
+    fn get_account(&self, account_id: &AccountId) -> Account {
+        self.accounts.get(account_id).unwrap_or_default()
+    }
+
+    fn save_account(&mut self, account_id: &AccountId, account: Account) {
+        if account != Account::default() {
```
ilblackdragon, line 279: take account: Option<Account> and on None delete?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+        let mut account = self.get_account(&account_id);
+
+        let num_shares = if self.total_staked_balance > 0 {
+            // price = total_staked / total_shares
+            // Price is fixed
+            // (total_staked + amount) / (total_shares + num_shares) = total_staked / total_shares
+            // (total_staked + amount) * total_shares = total_staked * (total_shares + num_shares)
+            // amount * total_shares = total_staked * num_shares
+            // num_shares = amount * total_shares / total_staked
+            // NOTE: Number of shares the account gets is rounded up, but the gas rebate within
```
ilblackdragon, line 162: hm, that's a bit dangerous IMO. what if governance removes rebates? why not round down?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+        env::panic(b"Staking contract should be initialized before usage")
+    }
+}
+
+#[near_bindgen]
+impl StakingContract {
+    /// Call to initialize the contract.
+    /// Specify which account can change the staking key and the initial staking key with ED25519 curve.
+    #[init]
+    pub fn new(owner_id: AccountId, stake_public_key: Base58PublicKey) -> Self {
```
ilblackdragon, line 67: should we add a fee? or let validators add it themself?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+            last_locked_account_balance: 0,
+            last_account_balance: env::account_balance(),
+            total_staked_balance: 0,
+            total_staked_shares: 0,
+            accounts: Map::new(b"u".to_vec()),
+        }
+    }
+
+    /// Call to update state after epoch switched.
+    pub fn ping(&mut self) {
```
ilblackdragon, line 84: so my understanding that there is no need in public ping function anymore?
e.g. you need to update internal state on any call, but don't need to keep pinging contract every epoch?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+        );
+        assert!(
+            account.unstaked_available_epoch_height <= env::epoch_height(),
+            "The unstaked balance is not yet available due to unstaking delay"
+        );
+        account.unstaked -= amount;
+        self.save_account(&account_id, account);
+        Promise::new(account_id).transfer(amount);
+
+        if self.total_staked_balance > self.last_locked_account_balance {
```
ilblackdragon, line 140: it's just weird that it's here and in `stake` given these variables only get updated in `ping`

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+        let mut account = self.get_account(&account_id);
+
+        assert!(
+            self.total_staked_balance > 0,
+            "The contract doesn't have staked balance"
+        );
+        // NOTE: The number of shares the account will pay is rounded up, to avoid giving extra
+        // amount.
+        // Rounding up `a / b` using `(a + b - 1) / b`.
+        let num_shares = ((U256::from(self.total_staked_shares) * U256::from(amount)
```
ilblackdragon, line 205: put shares computation into separate function as it's the same as line 165 and can have a nice comment there.
also can then test separately (in module level tests can set private fields directly).

staking-pool-shares/src/lib.rs
ilblackdragon: I guess it's when this function is called from deposit function.

evgenykuzyakov: Yep, in case it's called from deposit, the `account_balance` include the deposited amount

staking-pool-shares/src/lib.rs
evgenykuzyakov: This is easier to manage. It's basically clears the map when the account is default. But actually I need to only compare `unstaked` and `staked_shares` to 0, and ignore `epoch_height`

staking-pool-shares/src/lib.rs
evgenykuzyakov: It would mean the lockup contract might lose some tokens during stake/unstake call, which might break staking logic. Within lockup contract, we should either get the amount of tokens that we staked or not rely store internal values. Then there are no issue of losing some small amount of tokens during staking.

staking-pool-shares/src/lib.rs
evgenykuzyakov: In the next iteration

staking-pool-shares/src/lib.rs
evgenykuzyakov: Right. The only issue is to update view calls for balances to use proper virtual rewards.

staking-pool-shares/src/lib.rs
evgenykuzyakov: This is to address the issue when someone unstaked some amount in `T-3` with total stake of `X`. Now at epoch `T` the locked balance is equal to `X`, but the desired amount post 3 epochs of rewards is `X + Rewards`. Which means we have to restake `X + Rewards` to not miss out on the rewards.

staking-pool-shares/src/lib.rs
fckt: this is really nice change btw

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+            self.amount = self.amount + self.locked_amount - total_stake;
+            self.locked_amount = total_stake;
+            // Second function call action
+            self.update_context(staking(), 0);
+            self.contract.internal_after_stake();
+        }
+
+        pub fn skip_epochs(&mut self, num: EpochHeight) {
+            self.epoch_height += num;
+            self.locked_amount = (self.locked_amount * (100 + u128::from(num))) / 100;
```
fckt, line 591: Please, comment these numbers

fckt: Is that correct that node re-invest staking reward back to the locked_amount automatically?

fckt: And the reward per epoch is 1%, correct?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+        );
+        // Owner got 10% of the rewards
+        assert_eq_in_near!(
+            emulator.contract.get_account_staked_balance(owner()).0,
+            ntoy(10_000)
+        );
+
+        emulator.skip_epochs(10);
+        // Overriding rewards (another 100K reward)
+        emulator.locked_amount = deposit_amount + ntoy(100_000) + ntoy(100_000);
```
fckt, line 657: In skip_epochs you calculate 
`self.locked_amount = (self.locked_amount * (100 + u128::from(num))) / 100;`

But then it's overrides immediately. Whats' a point in this calculation?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+        self.last_epoch_height = epoch_height;
+
+        // New total amount (both locked and unlocked balances).
+        // NOTE: We need to subtract `attached_deposit` in case `ping` called from `deposit` call.
+        let total_balance =
+            env::account_locked_balance() + env::account_balance() - env::attached_deposit();
+        // Old total amount.
+        let last_total_balance = self.last_account_balance + self.last_locked_account_balance;
+
+        // TODO: Verify if the reward can become negative.
```
bowenwang1996, line 156: Currently it cannot. With slashing it can.

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+            self.total_staked_balance += remaining_reward;
+
+            // Now buying "stake" shares for the contract owner at the new shares price.
+            let num_shares = self.num_shares_from_amount_rounded_up(owners_fee);
+            if num_shares > 0 {
+                // Updating owner's inner account
+                let owner_id = self.owner_id.clone();
+                let mut account = self.get_account(&owner_id);
+                account.stake_shares += num_shares;
+                self.save_account(&owner_id, &account);
```
bowenwang1996, line 179: when do other accounts get updated?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+            .function_call(
+                b"internal_after_stake".to_vec(),
+                b"{}".to_vec(),
+                0,
+                INTERNAL_AFTER_STAKE_GAS,
+            );
+    }
+
+    /// Private method to be called after stake action to update inner balances.
+    pub fn internal_after_stake(&mut self) {
```
bowenwang1996, line 381: it is not private?

staking-pool-shares/src/lib.rs
evgenykuzyakov: Added

staking-pool-shares/src/lib.rs
evgenykuzyakov: It has to be called by the contract as an action, so it's private in sense of `current == predecessor`

staking-pool-shares/src/lib.rs
evgenykuzyakov: They get rewards from price increase

staking-pool-shares/src/lib.rs
evgenykuzyakov: These are fake reward numbers

staking-pool-shares/src/lib.rs
evgenykuzyakov: To skip 10 epochs

staking-pool-shares/src/lib.rs
evgenykuzyakov: Then we can assert here

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,802 @@
...
+    /// The total amount of shares. It should be equal to the total amount of shares across all
+    /// accounts.
+    pub total_stake_shares: Balance,
+    /// The total staked balance.
+    pub total_staked_balance: Balance,
+    /// The fraction of the reward that goes to the owner of the staking pool for running the
+    /// validator node.
+    pub reward_fee_fraction: RewardFeeFraction,
+    /// Persistent map from an account ID hash to the corresponding account.
+    pub accounts: Map<AccountHash, Account>,
```
bowenwang1996: Why do we need to use hash here instead of `AccountId`?

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,802 @@
...
+    pub fn multiply(&self, value: Balance) -> Balance {
+        (U256::from(self.numerator) * U256::from(value) / U256::from(self.denominator)).as_u128()
+    }
+}
+
+/// Interface for a voting contract.
+#[ext_contract(ext_voting)]
+pub trait VoteContract {
+    /// Votes on the given proposal_id.
+    fn vote(&mut self, proposal_id: ProposalId);
```
bowenwang1996: there should also be how much stake you vote

staking-pool-shares/src/lib.rs
bowenwang1996: yes

staking-pool-shares/src/lib.rs
bowenwang1996: So actually what is the point of calculating `remaining_reward`? It seems that it is not used.

staking-pool-shares/src/lib.rs
```diff
@@ -0,0 +1,802 @@
...
+    }
+
+    /********************/
+    /* Internal methods */
+    /********************/
+
+    /// Potentially restakes the total staked balance in case in the locked amount is smaller than
+    /// the desired staked balance due to unstake action in the past.
+    fn maybe_restake(&mut self) {
+        if self.total_staked_balance > self.last_locked_account_balance {
```
bowenwang1996: why would this not be true?

fckt:
We need to check the case when the account.locked is decreasing (due to penalties) (writing it just to not forget).

bowenwang1996:
> when the account.locked is decreasing (due to penalties) 

It is not possible short term. We can revisit this later.

fckt:
Suppose validator (poll owner) wants to terminate his service. How it can be done?

evgenykuzyakov:
> Just think of a (unlikely)scenario: if someone delegates their tokens and then deleted their account, whoever recreates the account can claim the delegated tokens.

This makes sense. Checkout betanet. There are free tokens there if you can find them

staking-pool-shares/src/lib.rs
evgenykuzyakov: Obsolete

staking-pool-shares/src/lib.rs
evgenykuzyakov: To avoid slowing lookups with account collisions. Since the account ID is limited to 64 characters, the most you can slow down the lookup is by 64 nodes. So it's not critical to do this for account ID. Still, it's the best practice to hash user-input when storing internally, to avoid creating deep tries.

staking-pool-shares/src/lib.rs
evgenykuzyakov: Will add this

evgenykuzyakov:
> We need to check the case when the account.locked is decreasing (due to penalties) (just a memo).

When slashing is introduced, the staking contract might arrive into an invalid state, because the price of a share will fall below 1. This is an invariant for this implementation, so we don't need to handle it in the current implementation.

evgenykuzyakov:
> Suppose validator (pool owner) wants to terminate the service. How it can be done?

As discussed. You can't kick out existing users, so the only way is to shut down the node on owner's end and the contract will not be able to produce rewards. It's not punishable right now.

I also think every staking pool will have a public facing website where they can communicate to the delegates.

staking-pool/README.md
```diff
@@ -1,11 +1,122 @@
...
+The contract maintains information per delegator in the persistent map keyed by the hash of the delegator account ID.
+
+The information contains:
+- Unstaked balance of the account.
+- Number of "Stake" shares.
+- The minimum epoch height when the unstaked balance can be withdrawn.
+
+A delegator can do the following actions:
+
+**Deposit**
```
fckt, line 51: Maybe format as headings

staking-pool/src/lib.rs
```diff
@@ -90,358 +111,514 @@ impl Default for StakingContract {
...
+    /// Owner's method.
+    /// Vote on a given proposal with the given stake amount on a given voting contract account ID
+    /// on behalf of the pool.
+    /// NOTE: This method allows the owner to call `vote(proposal_id: U64, stake: U128)` on any
+    /// contract on behalf of this staking pool.
+    pub fn vote(
+        &mut self,
+        voting_account_id: AccountId,
+        proposal_id: ProposalId,
+        stake: U128,
```
fckt, line 468: What is stake? How it's verified?

evgenykuzyakov: Should be part of voting spec: https://github.com/nearprotocol/NEPs/pull/62

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+# Staking / Delegation contract
 
-*This is an experimental contract. Please use only on TestNet.*
-
-This contract provides a way for other users to delegate funds to a single staker.
+This contract provides a way for other users to delegate funds to a single validation node.
 
 Implements the https://github.com/nearprotocol/NEPs/pull/27 standard.
 
+There are 3 different roles:
```
willemneal, line 7: Nit, but numbers ten and under are typically spelled out.

willemneal, line 7:
```suggestion
There are three different roles:
```

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+- The staking pool contract account `my_validator`. A key-less account with the contract that pools funds.
+- The owner of the staking contract `owner`. Owner runs the validator node on behalf of the staking pool account.
+- A delegator account `user1`. The account who wants to stake their fund to the pool.
```
willemneal, line 10: Why not `user`?

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+
+The owner can setup such contract and validate on behalf of this contract in their node.
+Any other user can send their tokens to the contract, which will be pooled together and increase the total stake.
+These users would accrue rewards (subtracted fees set by the owner).
```
willemneal, line 14:
```suggestion
These users accrue rewards (subtracted fees set by the owner).
```

willemneal: Also it's not clear that they must deposit and then stake the tokens.

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+Then they can unstake and withdraw their balance after some unlocking period.
+
+## Staking pool implementation details
+
+For secure operation of the staking pool, the contract should not have any access keys.
```
willemneal, line 19: Need to explain why.

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+
+This staking contract pools together staked tokens and issues "stake" shares to the owners of these tokens.
```
willemneal, line 21:
```suggestion
After users deposit tokens to the contract, they can stake some or all of them to receive shares.
```
Or something similar.

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+A contract has to be initialized with the following parameters:
+- `owner_id` - `string` the account ID of the contract owner. This account will be able to call owner-only methods. E.g. `owner`
+- `stake_public_key` - `string` the initial public key that will be used for staking on behalf of the contract's account in base58 ED25519 curve. E.g. `KuTCtARNzxZQ3YvXDeLjx83FDqxv2SdQTSbiq876zR7`
+- `reward_fee_fraction` - `json serialized object` the initial value of the fraction of the reward that the owner charges delegators for running the node.
+The fraction is defined by the numerator and denumerator with `u32` types. E.g. `{numerator: 10, denominator: 100}` defines `10%` reward fee.
+The fraction can be at most `1`. The denumerator can't be `0`.
+
+During the initialization the contract checks validity of the input and initializes the contract.
+The contract shouldn't have locked balance during the initialization.
+All current balance is converted to shares and will be staked (after the next action).
```
willemneal, line 36:
```suggestion
The current total balance is converted to shares and will be staked (after the next action).
```

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+This balance can never be unstaked or withdrawn from the contract.
+It's used to maintain the minimum number of shares, as well as help pay for the potentially growing contract storage.
+
+### Delegator accounts
+
+The contract maintains information per delegator in the persistent map keyed by the hash of the delegator account ID.
+
+The information contains:
+- Unstaked balance of the account.
+- Number of "Stake" shares.
```
willemneal, line 46:
```suggestion
- Number of "stake" shares.
```

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+
+During the initialization the contract checks validity of the input and initializes the contract.
+The contract shouldn't have locked balance during the initialization.
+All current balance is converted to shares and will be staked (after the next action).
+This balance can never be unstaked or withdrawn from the contract.
+It's used to maintain the minimum number of shares, as well as help pay for the potentially growing contract storage.
+
+### Delegator accounts
+
+The contract maintains information per delegator in the persistent map keyed by the hash of the delegator account ID.
```
willemneal, line 42:
```suggestion
The contract maintains account information per delegator associated with the hash of the delegator's account ID.
```

staking-pool/src/lib.rs
```diff
@@ -90,358 +109,513 @@ impl Default for StakingContract {
...
+        );
+
+        if need_to_restake {
+            self.restake();
+        }
     }
 
     /// Withdraws the non staked balance for given account.
+    /// It's only allowed if the `unstake` action was not performed in the recent 3 epochs.
     pub fn withdraw(&mut self, amount: U128) {
```
willemneal, line 212: When an account has withdraw all of their tokens, shouldn't it be deleted from the map?

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+
+The information contains:
+- Unstaked balance of the account.
+- Number of "Stake" shares.
+- The minimum epoch height when the unstaked balance can be withdrawn.
```
willemneal, line 47:
```suggestion
- The minimum epoch height when the unstaked balance can be withdrawn. Initially zero.
```

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+
+A delegator can do the following actions:
+
+#### Deposit
+
+When a delegator account first deposit funds to the contract, the internal account is created and credited with the
```
willemneal, line 53:
```suggestion
When a delegator account first deposits funds to the contract, the internal account is created and credited with the
```

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+### Reward distribution
+
+Before every action the contract calls method `internal_ping`.
+This method distributes rewards towards active delegators when the blockchain epoch switches.
+The rewards might be given due to staking and also because the contract earns gas fee rebates for every function call.
+
+The method first checks that the current epoch is different from the last epoch, and if it's not changed exits the method.
+
+The reward are computed the following way. The contract keeps track of the last known total account balance.
+This balance consist of the initial contract balance, and all delegator account balances (including the owner) and all accumulated rewards.
```
willemneal, line 88: >  and all accumulated rewards.

How are they added to the contract?

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+
+#### Stake
+
+When an account wants to stake a given amount, the contract calculates the number of "stake" shares (`num_shares`) and the actual rounded stake amount (`amount`).
+The unstaked balance of the account is decreased by `amount`, the number of "stake" shares of the account is increased by `num_shares`.
+The contract increases the total number of staked tokens and the total number of "stake" shares. Then the contract restakes.
+
+#### Unstake
+
+When an account wants to unstake a given amount, the contract calculates the number of "stake" shares needed (`num_shares`) and
```
willemneal, line 64: Mention that that this is calculated with the current price of stake share.

staking-pool/README.md
```diff
@@ -1,11 +1,131 @@
...
+- Vote on behalf of the pool. This is needed for the NEAR chain governence, and can be discussed in the following NEP: https://github.com/nearprotocol/NEPs/pull/62
+
+## Staking pool contract guarantees and invariants
+
+This staking pool implementation guarantees the required properties of the staking pool standard:
+
+- The contract can't lose or lock tokens of users.
+- If a user deposited X, the user should be able to withdraw at least X.
+- If a user successfully staked X, the user can unstake at least X.
+- The contract should lock unstaked funds for longer than 4 epochs.
```
willemneal, line 117: Shouldn't this be the opposite?

staking-pool/README.md
bowenwang1996: Automatically.

staking-pool/README.md
bowenwang1996: What do you mean by opposite?

willemneal: I mean for no more than 4.  Should it ever lock for 5 or more?

staking-pool/README.md
willemneal: Can you elaborate?  If I was making a test how would this be done?  Is it a special transaction?

staking-pool/README.md
evgenykuzyakov: Updated.

staking-pool/README.md
evgenykuzyakov: @willemneal You're correct about "should not"

staking-pool/Cargo.toml
```diff
@@ -1,26 +1,37 @@
 [package]
-name = "staking-contract"
+name = "staking-pool"
 version = "0.1.0"
-authors = ["Illia Polosukhin <ilblackdragon@gmail.com>"]
+authors = ["Near Inc <hello@nearprotocol.com>"]
```
willemneal, line 4:
```suggestion
authors = ["Near Inc <hello@near.org>"]
```
Does this apply here too?

staking-pool/README.md
fckt: account.locked just increasing from time to time. That increase is a revard (inflation). But the amount of shares is fixed to the total locked balance as if there was no inflation.

PR #12, opened 2020-05-12
Adding basic multisig contract

ilblackdragon:
Logic:
 - Create contract with N function access keys.
 - Request to transfer / add keys / function calls
 - Need K confirm calls for given request for it to execute.

TODO: more detailed deployment and usage documentation.


multisig/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+        env::panic(b"Multisig contract should be initialized before usage")
+    }
+}
+
+#[near_bindgen]
+impl MultiSigContract {
+    /// Initialize multisig contract.
+    /// @params num_confirmations: k of n signatures required to perform operations.
+    #[init]
+    pub fn new(num_confirmations: u32) -> Self {
```
evgenykuzyakov, line 50: You may want to initialize keys here by passing them and then adding them.

multisig/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+use near_sdk::collections::Map;
+use near_sdk::json_types::{Base58PublicKey, U128};
+use near_sdk::{env, near_bindgen, AccountId, Balance, Gas, Promise, PublicKey};
+
+const DEFAULT_ALLOWANCE: u128 = 1_000_000_000_000_000_000_000;
+
+pub type RequestId = u32;
+
+#[derive(BorshDeserialize, BorshSerialize)]
+pub enum MultiSigRequest {
```
evgenykuzyakov, line 13: May want to remove keys

evgenykuzyakov, line 13: And to adjust the `K` out of `N`

multisig/src/lib.rs
ilblackdragon: This should be done via transaction (started writing that in readme how to do that)

ilblackdragon: The suboptimal part is that there is no way to check all access keys from contract.

multisig/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+            method_name,
+            args,
+            deposit,
+            gas,
+        })
+    }
+
+    /// Confirm given request with given signing key.
+    /// If with this, there has been enough confirmation, a promise with request will be scheduled.
+    pub fn confirm(&mut self, request_id: RequestId) {
```
evgenykuzyakov, line 198: I think you should return `PromiseOrValue<bool>` here, when you execute the request, you return the promise. When adding a new confirmation, return `true`.

multisig/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+            "No such request: either wrong number or already confirmed"
+        );
+        assert!(
+            self.confirmations.get(&request_id).is_some(),
+            "Internal error: confirmations mismatch requests"
+        );
+        let mut confirmations = self.confirmations.get(&request_id).unwrap();
+        assert!(
+            !confirmations.contains(&env::signer_account_pk()),
+            "Already confirmed this request"
```
evgenykuzyakov, line 215:
```suggestion
            "Already confirmed this request with this key"
```

multisig/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+pub type RequestId = u32;
+
+#[derive(BorshDeserialize, BorshSerialize)]
+pub enum MultiSigRequest {
+    Transfer {
+        account_id: AccountId,
+        amount: Balance,
+    },
+    AddKey {
+        public_key: PublicKey,
```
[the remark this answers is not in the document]
ilblackdragon: why? the public functions taking Base58PublicKey but this is internal data structure.

multisig/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+use near_sdk::collections::Map;
+use near_sdk::json_types::{Base58PublicKey, U128};
+use near_sdk::{env, near_bindgen, AccountId, Balance, Gas, Promise, PublicKey};
+
+/// Unlimited allowance for multisig keys.
+const DEFAULT_ALLOWANCE: u128 = 0;
+
+pub type RequestId = u32;
+
+#[derive(BorshDeserialize, BorshSerialize)]
```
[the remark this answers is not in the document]
ilblackdragon: got it https://github.com/near/initial-contracts/pull/12#discussion_r430660497

multisig/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+            MultiSigRequest::Transfer {
+                receiver_id,
+                amount,
+            } => PromiseOrValue::Promise(Promise::new(receiver_id).transfer(amount.into())),
+            MultiSigRequest::AddKey { public_key } => PromiseOrValue::Promise(
+                Promise::new(env::current_account_id()).add_access_key(
+                    public_key.into(),
+                    DEFAULT_ALLOWANCE,
+                    env::current_account_id(),
+                    "transfer,function_call,add_key,confirm"
```
evgenykuzyakov, line 113: This needs to be updated

multisig/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+        self.confirmations.remove(&request_id);
+    }
+
+    fn execute_request(&mut self, request: MultiSigRequest) -> PromiseOrValue<bool> {
+        match request {
+            MultiSigRequest::Transfer {
+                receiver_id,
+                amount,
+            } => PromiseOrValue::Promise(Promise::new(receiver_id).transfer(amount.into())),
+            MultiSigRequest::AddKey { public_key } => PromiseOrValue::Promise(
```
evgenykuzyakov, line 108: Instead of `PromiseOrValue::Promise(...)` you can use `Promise::new(...).into()`

multisig/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+            "Already confirmed this request with this key"
+        );
+        if confirmations.len() as u32 + 1 >= self.num_confirmations {
+            let request = self
+                .requests
+                .remove(&request_id)
+                .expect("Failed to remove existing element");
+            let result = self.execute_request(request);
+            self.confirmations.remove(&request_id);
+            result
```
evgenykuzyakov, line 168:
```suggestion
            self.confirmations.remove(&request_id);
            self.execute_request(request)
```

multisig/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+    /// Initialize multisig contract.
+    /// @params num_confirmations: k of n signatures required to perform operations.
+    #[init]
+    pub fn new(num_confirmations: u32) -> Self {
+        assert!(!env::state_exists(), "Already initialized");
+        Self {
+            num_confirmations,
+            request_nonce: 0,
+            requests: Default::default(),
+            confirmations: Default::default(),
```
evgenykuzyakov, line 64: I prefer to explicitly give prefixes for the collections to avoid conflicts if later you have to remove something or change the order.

```suggestion
            requests: Map::new(b"r".to_vec()),
            confirmations: Map::new(b"c".to_vec()),
```

PR #19, opened 2020-05-16
Updated README

htafolla:
several commands were missing option --accountId


htafolla:
near-shell was patched and param is no longer needed.

PR #33, opened 2020-05-23
feat: add getter for stake public key

bowenwang1996:
Add a view method to query the stake public key of the staking pool. This is useful for validators to double check whether they have used the right key.


staking-pool/src/lib.rs
```diff
@@ -426,6 +426,11 @@ impl StakingContract {
         self.reward_fee_fraction.clone()
     }
 
+    /// Returns the staking public key
+    pub fn get_stake_public_key(&self) -> Base58PublicKey {
+        Base58PublicKey(self.stake_public_key.clone())
```
evgenykuzyakov, line 431:
```suggestion
        self.stake_public_key.clone().try_into().unwrap()
```

bowenwang1996:
@evgenykuzyakov done

bowenwang1996:
@evgenykuzyakov updated

PR #37, opened 2020-05-26
Staking pool factory

evgenykuzyakov:
This contract allows to deploy new staking pool contract and automatically whitelist them.

- [x] Add a few more integration tests


staking-pool-factory/README.md
```diff
@@ -0,0 +1,57 @@
+# Staking Pool Factory Contract
+
+This contract allows to deploy new staking pool contract and automatically whitelist them.
```
willemneal, line 3:
```suggestion
This contract deploys and automatically whitelists new staking pool contracts.
```

staking-pool-factory/README.md
```diff
@@ -0,0 +1,57 @@
...
+It allows any user to create an new whitelisted staking pool.
+
+The staking pool factory contract keeps the hardcoded binary of the staking pool contract within itself.
```
willemneal, line 6:
```suggestion
The staking pool factory contract packages the binary of the staking pool contract within its own binary.
```

staking-pool-factory/README.md
```diff
@@ -0,0 +1,57 @@
...
+To create a new staking pool a user should issue a function call transaction and attach the required minimum deposit.
+The entire deposit will be transferred to the newly created staking pool contract in order to cover the required storage.
+
+When a user issues a function call towards the factory to create a new staking pool the factory internally checks that
```
willemneal, line 10: What does it mean to internally check?  Does this mean that it checks that the account ID doesn't exist as an account on chain?

staking-pool-factory/README.md
```diff
@@ -0,0 +1,57 @@
...
+the staking pool account ID doesn't exists, validates arguments for the staking pool initialization and then issues a
```
willemneal, line 11: nit
```suggestion
the staking pool account ID does not exists, validates arguments for the staking pool initialization and then issues a
```

staking-pool-factory/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize)]
+pub struct StakingPoolFactory {
+    /// Account ID of the staking pool whitelist contract.
+    staking_pool_whitelist_account_id: AccountId,
+
+    /// The account ID of the staking pools created.
+    staking_pool_account_ids: Set<AccountId>,
```
willemneal, line 45: Isn't this information stored in the whitelist contract?

willemneal:
LGTM just small nits no README and one question about design.

staking-pool-factory/src/lib.rs
evgenykuzyakov: It's different, here it mostly to avoid creation of the staking pools with the same name.

willemneal: Yeah, but can't you make a promise call to the whitelist to do that?

evgenykuzyakov: That's extra call, but in theory nothing breaks if we don't maintain the `Set` internally. It just saves some gas if 2 people try to create the same pool. Then one of the creations is going to fail.

willemneal: Okay makes sense to me.  Then everything LGTM.

PR #43, opened 2020-05-30
Voting contract for unlocking transfer

bowenwang1996:
The purpose of this contract is solely for validators to vote on whether to unlock
token transfer. Validators can call `vote` to vote for yes with the amount of stake they wish
to put on the vote. If there are more than 2/3 of the stake at any given moment voting for yes, the voting is done. After the voting is finished, no one can further modify the contract.


voting/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+            self.result.is_none(),
+            "check result is called after result is already set"
+        );
+        let total_stake = env::validator_total_stake();
+        if self.total_voted_stake > 2 * total_stake / 3 {
+            self.result = Some(env::block_timestamp());
+        }
+    }
+
+    /// Vote on a specific proposal with the given stake
```
ilblackdragon, line 89: no proposals any more, right? the comment seems wrong.

Also should we even vote with stake amount? If we don't have ability for validators to proxy their delegators votes?

Should it just be that validator can call vote, when they think it's ready to unlock?
And it records and checks their stake at the time of `check_results` call? 
This way I can vote right away, but if people don't think I'm right - they will undelegate to someone else.

This will also be much simpler, because wouldn't need this `resolve_votes` magic.

voting/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+            self.account_stake.remove(&account_id);
+            return;
+        }
+        self.votes.insert(&account_id, &stake);
+        self.account_stake.insert(&account_id, &account_stake);
+        self.total_voted_stake = self.total_voted_stake + stake - voted_stake;
+        self.check_result();
+    }
+
+    pub fn get_result(&self) -> Option<Timestamp> {
```
ilblackdragon, line 131: hm, is Option fine for lockup and all other contracts?
should we just set a backup date that it will return or if the stake voted - it will update it to closer date?

voting/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+                    self.account_stake
+                        .insert(&account_id, &account_current_stake);
+                }
+            }
+            self.check_result();
+            self.last_epoch_height = cur_epoch_height;
+        }
+    }
+
+    fn check_result(&mut self) {
```
ilblackdragon, line 78: this should be public, because with the same votes - it can pass later when total stake decreases.

voting/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
ilblackdragon, line 1: as I said in prev PR - license should live on the outside and should probably copy one we use everywhere else.

voting/src/lib.rs
bowenwang1996: @evgenykuzyakov I thought it is fine to use `Option` here?

voting/src/lib.rs
bowenwang1996: I think this sounds reasonable. @evgenykuzyakov what do you think?

voting/src/lib.rs
bowenwang1996: I made `ping` public.

voting/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /// Ping to update the votes according to current stake of validators.
+    pub fn ping(&mut self) {
+        assert!(self.result.is_none(), "Voting has already ended");
+        let cur_epoch_height = env::epoch_height();
+        if cur_epoch_height != self.last_epoch_height {
+            for account_id in self.votes.keys().into_iter().collect::<Vec<_>>() {
+                let account_current_stake = env::validator_stake(&account_id);
+                let account_voted_stake = self.votes.remove(&account_id).unwrap();
```
ilblackdragon, line 50: Self.votes.get?
After remove the votes will be gone?
Also why does this has a value? Should this just be a hashset of Validators that voted?

bowenwang1996: Good point. I am retarded.

bowenwang1996: Actually it's needed when a validator who already voted calls `vote` again. If we don't store their voted stake we don't know how to update `total_voted_stake`.

voting/src/lib.rs
evgenykuzyakov: Yes, option is expected here. I never heard about backup date?

voting/src/lib.rs
```diff
@@ -0,0 +1,284 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use near_sdk::collections::Map;
+use near_sdk::{env, near_bindgen, AccountId, Balance, EpochHeight};
+
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+type Timestamp = u64;
```
evgenykuzyakov, line 8: Let's change to `U64` and call it `WrappedTimestamp` to be json friendly. Otherwise modify the actual returned value in `get_result` to be `Option<U64>`
```suggestion
type WrappedTimestamp = U64;
```

voting/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+version = "0.1.0"
+authors = ["Near Inc <hello@nearprotocol.com>"]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0.45"
```
evgenykuzyakov, line 12:
```suggestion
serde_json = "1.0"
```

voting/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        self.total_voted_stake = self.total_voted_stake + account_stake - voted_stake;
+        if account_stake > 0 {
+            self.votes.insert(&account_id, &account_stake);
+            self.check_result();
+        }
+    }
+
+    /// Vote for unlocking transfer
+    pub fn vote(&mut self) {
+        self.vote_internal(true)
```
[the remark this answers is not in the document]
bowenwang1996: As @ilblackdragon pointed out, there seems to be no point in allowing partial votes. If you don't agree with some pool's voting you can just unstake and leave.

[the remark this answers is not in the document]
evgenykuzyakov: Ok, works for me. Should we merge `vote/withdraw` into `vote(yes/no)`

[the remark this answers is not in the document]
bowenwang1996: Ah yes that is better for the staking pool. I forgot.

voting/src/lib.rs
ilblackdragon: well, that was a suggestion if it makes it easier. We don't need it.

Issue #48, opened 2020-06-02
Removal of all Main PublicKeys


nearmax:
Impact: High
Likelihood: Low
Reported by a third-party.

**Description**
In the lockup contract, main public keys are used by the owner to operate certain owner functions such as adding and removing other keys and transferring funds.

The current owner is able to remove all keys including their own key, leaving the contract with no remaining main keys. This would effectively lock any remaining balance in the contract.

An accidental case where this occurs would be when a user sends two transactions, one to add a new key and the other to remove the existing key. If these two transactions are executed in the wrong order (i.e. the deletion first). The contract will be locked.

**Recommendations**
To add a level of protection to accidental deletion it is recommended to either:
• Prevent deletion of the signing key;
• Ensure there is at least one main key in the list of keys

ilblackdragon:
Related - https://github.com/nearprotocol/NEPs/issues/79

PR #52, opened 2020-06-03
Prevent deletion of the signing key

evgenykuzyakov:
Fixes: #48 

Don't let delete signer public key. This will prevent from deleting all public keys and locking the account.


lockup/src/lib.rs
```diff
@@ -295,6 +295,19 @@ mod tests {
         contract.add_staking_access_key(public_key(4));
     }
 
+    #[test]
+    #[should_panic(expected = "Can not remove signer public key")]
```
fckt, line 299:
```suggestion
    #[should_panic(expected = "Can not remove owner public key")]
```

evgenykuzyakov: It still need to emphasize why you can't remove owner's public key. Basically, saying the key used to sign the transaction.

evgenykuzyakov: Updated to `Can not remove owners public key that is used to sign this transaction`

Issue #58, opened 2020-06-03
Miscellaneous General Comments


nearmax:
Informational
Reported by a third party.

This section details miscellaneous findings identified by the testing team that do not have a direct security implication:
**fungible-token** 
- [x]  line [56] of fungible-token/src/lib.rs includes a comment with no message. (i.e. // ).
- [x]  It should be documented that the contract account should be keyless or there is the potential for modification by the owner.

**staking-pool**
- [x] There is no limit to the owner’s reward percentage. They may modify it to 100% and claim all rewards.
- [x]  Any transfers made by a user (i.e. using send ) without a function call will be treated as rewards and distributed to stakers accordingly.
- [x]  If an account calls withdraw() and the same account is deleted before the transfer() promise
has been executed, the transfer will fail but the changes to StakingContract will not be reverted.
The result is that the fees will be distributed to the remaining stakers as rewards during the next
internal_ping() .
- [x]  If a validator is slashed, the remaining funds will be locked in the contract until someone sends a
sufficient balance for the contract to have total_balance >= last_total_balance .
- [x]  line [27]: /// A type to distinguish between a balance and a staking shares for better
readability. -> and staking shares
- [x]  line [32]: The to do, // TODO: Revert back to 4 once wasm/wasmer bug is fixed. , may now
be updated.
- [x]  line [72]: Which will not unlock the funds for 4 epochs. ->
It will not unlock the funds for 4 epochs.
- [x]  line [146]: /// It prevents inflation the price of the share too much. ->
/// It prevents inflation of the price of the share too much.
- [x]  line [214]: /// It’s only allowed if the ‘unstake‘ action was not performed in the recent
3 epochs. -> in the 3 most recent epochs.
- [x]  line [275] & line [346]: Calculated staked amount be positive, ... ->
Calculated staked amount must be positive, ...
- [x]  line [383]: https://github.com/nearprotocol/nearcore/issues/2636 , the associated PR has
been merged and note can be removed.
- [x]  line [495]: // Checking if we need there are rewards to distribute. ->
// Checking, if we need, there are rewards to distribute.
- [x]  line [520]: // Now buying "stake" shares for the contract owner at the new shares price.
-> at the new share price.
- [x]  line [532]: // got any shares or not. -> has any shares or not.
- [x]  line [617]: /// Inner method to get the save the given account for a given account ID.
-> Inner method to save the ...

**lockup**
- [x] Type ProposalId is u64 as opposed to U64 .
[...]

PR #60, opened 2020-06-04
Lockup misc comments and update PollResult

evgenykuzyakov:
Update `PollResult` to be simple `Option<U64>`.

Addresses comments from https://github.com/near/core-contracts/issues/58


lockup/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! A smart contract that allows tokens lockup.
+//! A smart contract that allows tokens locked up.
```
willemneal, line 1:
```suggestion
//! A smart contract that allows tokens to be locked up.
```

Issue #40, opened 2020-05-27
Docker image to build the contracts


nearmax:
We need to create a docker image (Docker file is enough, no need to publish it) that people can use to build this contracts. This Docker image will have fully fixed versions of the toolchain which will guarantee 100% reproducibility of the Wasm files that it outputs given the source code. This will allow community members to verify that the Wasm files that were deployed were in fact compiled from the provided source code.

PR #64, opened 2020-06-04
docker builder for building contracts

ailisp:
#40 Also update all wasm to be built by this docker. By building on docker I found it's conflicting of rust-toolchain file and cargo +stable (error: stable is not installed). Since we want pinned toolchain, I think we should all stick with version in rust-toolchain file.


ailisp:
We need also pin to a specific stable rust version to produce deterministic wasm output. We now pin rust-toolchain to nearcore's rust-toolchain, but that's a nightly and for use near-runtime-standalone in tests. Does latest (1.44.1) sounds good?

ailisp:
@olonho a working contract builder image, might be helpful to you

PR #69, opened 2020-06-12
Single tx multisig with multiple actions

mattlockyer:
Based on:
997dfc1

Allow each request (transaction) to have multiple actions.
Adds CreateAccount action.


mattlockyer:
A front-running style attack:
- someone gets 1/3 keys
- listens for `add_request`
- calls `delete_request` anytime a request is added

This essentially renders the multisig useless, provided the attacker is able to keep up attack.

mattlockyer:
> This essentially renders the multisig useless, provided the attacker is able to keep up attack.
Yes we can always call `add_request` and `confirm` from one access key immediately, but we must assume some real world time passes before subsequent confirmations. This is the attackers opportunity.
The only mitigation I see is restricting who can call `delete_request`.

multisig/README.md
```diff
@@ -19,18 +19,28 @@ All operations going forward will require `K` signatures to be performed.
...
         method_name: String,
         args: Base64VecU8,
         deposit: U128,
         gas: Gas
-    },
- SetNumConfirmations { num_confirmations: u32 }
-``` 
+  },
+  SetNumConfirmations { num_confirmations: u32 },
+  CreateAccount,
```
ilblackdragon, line 33: let's also add DeployContract from that PR

multisig/README.md
```diff
@@ -54,6 +64,8 @@ Guarantees:
 ### Gotchas
  
 User can delete access keys such that total number of different access keys will fall below `num_confirmations`, rendering contract locked.
+
+`AddKey` on another account adds full access key.
```
ilblackdragon, line 68: Actually in the other PR added optional permissions to allow non full access keys

ilblackdragon:
I agree on expiration, but 5 minutes is way too little.
This is general multisig, and I'm planning to use it across many people for example.
May be let's add "timeout" into `add_request` which sets how long a request is valid.

There is also a possible issue around gas usage of the keys (see advisory in the repo)

mattlockyer:
@ilblackdragon @evgenykuzyakov please have a look at new fields `request_expiry` and `request_lifespan` and the multisig actions to set them.

`request_lifespan`: prevents against malicious person obtaining 1 of the keys and calling `delete_request` before confirmations can be attempted. Requests still within lifespan cannot be deleted.

`request_expiry`: prevents against malicious person flooding contract with requests and / or user accidentally confirming old requests. Expired requests are removed.

https://github.com/near/core-contracts/pull/69/commits/bd39da97109af3a002b8ba59cc003fef756fd4c1

multisig/src/lib.rs
```diff
@@ -107,22 +134,16 @@ impl MultiSigContract {
...
-            "No such request: either wrong number or already confirmed"
-        );
-        assert!(
-            self.confirmations.get(&request_id).is_some(),
-            "Internal error: confirmations mismatch requests"
-        );
-        self.requests.remove(&request_id);
-        self.confirmations.remove(&request_id);
+    pub fn delete_request(&mut self, request_id: RequestId) -> bool {
+        self.valid_request(request_id);
```
ilblackdragon, line 138: may be then `let added = self.get_valid_request_or_fail(request_id).added` ?

because actually there is no reason to do `assert!(
            self.requests.get(&request_id).is_some(),` given `self.requests.get(&request_id).expect(..)` will do the same.

multisig/src/lib.rs
```diff
@@ -196,6 +208,34 @@ impl MultiSigContract {
                     deposit.into(),
                     gas.into(),
                 ),
+                // the following must be a single action request with receiver_id == contract_id
+                MultiSigRequestAction::SetNumConfirmations { num_confirmations } => {
+                    assert_eq!(request.receiver_id, env::current_account_id(), "Changing number of confirmations only works when receiver_id is equal to current_account_id");
```
ilblackdragon, line 213: can also wrap into a helper function check_single_action("Changing number ...")

mattlockyer:
> Just use `u64` and nanoseconds. Should be easier to handle within contract than think about rounding

Removed time checking altogether. I feel it's the wrong approach.

mattlockyer:
@ilblackdragon @evgenykuzyakov 
I've removed the time checking altogether.

- Added a `signer_pk` to each request in `struct MultiSigRequestWithSigner`.
- Incrementing `num_requests_pk` for every `add_request`; decrementing on `delete_request` and when confirmations are reached.
- Added helper for removing request, removing confirmations and decrementing `num_requests_pk` based on `signer_pk` of `MultiSigRequestWithSigner`.

Only the original request creator can call `delete_request`, but any key can reach confirmations and request is removed.

There is an `active_requests_limit: 4` default for now. Should we make it configurable?

vgrichina:
>  I would have to bring back some notion of time checking to prevent the attack from happening too quickly.

@mattlockyer I think we need to have a limit on number of unconfirmed requests per unit of time per given key. Say no more than 10 requests without confirmation per day.

When the limit is hit it also makes sense if our 2FA service sends out a warning explaining what might be happening and what to do if somebody is actively trying to compromise your account.

ilblackdragon:
@mattlockyer @vgrichina @evgenykuzyakov I want to schedule post mortem for our failure to design this well from the start. And there let's finalize the use cases and design.

multisig/README.md
ilblackdragon: Let's also port the comments and other stuff from here: https://github.com/near/core-contracts/pull/66/files#diff-6620ce7e2df548737fd0e98c913fcc8aR22

mattlockyer:
I'm ok with the multisig as is right now.
RE: upgrades
I'm not expert enough to say right now if this will be really difficult.
RE: multiple txs
Considering the infra isn't built on the client side, and likely won't be there for some time, I think it's ok to keep the contract simple. Especially given that it's much easier to tell which tx failed when we chain them in the client. We can also set up the wallet (in the future) to add multiple requests and then ask for a single 2fa confirmation on an array of `request_id` that we chain confirm in `near-contract-helper` and handle errors as they occur.

I re-added a time restriction to delete_request, so you can make at most 12 requests in a 12 hour period. This should prevent a contract from being drained too quickly, 1 malicious request / hr at most.

@vgrichina You can also get the number of requests for a given key on your account and check for a high number of active requests that are not confirmed, this tells you something is wrong.
@ilblackdragon added the comments you requests, did I miss anything?

mattlockyer:
@evgenykuzyakov @ilblackdragon I think we're down to security.
RE: https://github.com/near/core-contracts/security/advisories/GHSA-hxfc-5rpv-frqj
I removed a mitigation against repeated `add_request` `delete_request` gas burn attack after discussing with @vgrichina.  We felt it should be handled by allowance.
However, to limit all storage being used by too many `add_request` and locking all tokens from gas, there is still `active_requests_limit` enforced per key.

The downside to using allowance to mitigate a gas burn attack on the contract is that allowances must be managed (topped up) for valid keys at some point in the future by deleting the access key and adding a new one.
https://nomicon.io/DataStructures/AccessKey.html?highlight=allowance#accesskeypermissionfunctioncall

Are we missing any helper functions for access key rotation that would make the dev (wallet developer) experience better?

vgrichina:
> Are we missing any helper functions for access key rotation that would make the dev (wallet developer) experience better?

@mattlockyer I think one potential workaround is just to remove key and add it back with topped up allowance every time request is successfully approved.

mattlockyer:
> @mattlockyer I think one potential workaround is just to remove key and add it back with topped up allowance every time request is successfully approved (this can be atomic multi-action tx).

I like it, but where would it happen? If directly in the confirm method of the contract itself, we can technically do it but a couple of things to consider:
- how much gas is delete_key, add_key?
- does this introduce any security issues?
- make sure allowance is enough to do the original multisig request actions AND this key refresh

vgrichina:
> * how much gas is delete_key, add_key?
> * make sure allowance is enough to do the original multisig request actions AND this key refresh

you can try it and look up in block explorer, but generally it shouldn't be a lot (see e.g. https://explorer.testnet.near.org/transactions/7USmNoXycwifyM3JN1swkmgkVjTLh97tkzR5jDAUoLy8) for creating account and adding key

> * does this introduce any security issues?

doesn't seem to be an issue given that multisig request is already confirmed at the moment (so user is assumed to be legit)

multisig/src/lib.rs
```diff
@@ -64,110 +100,147 @@ impl MultiSigContract {
...
+            promise = match action {
+                MultiSigRequestAction::Transfer { amount } => promise.transfer(amount.into()),
+                MultiSigRequestAction::CreateAccount => promise.create_account(),
+                MultiSigRequestAction::DeployContract { code } => {
+                    promise.deploy_contract(code.into())
+                }
+                MultiSigRequestAction::AddKey {
+                    public_key,
+                    permission,
+                } if request.receiver_id == env::current_account_id() => {
```
evgenykuzyakov, line 168: I think you should be able to add `FunctionCallPermission` access keys to the multi-sig. We just need to make sure that `permission.receiver_id` doesn't equal to `env::current_account_id()`. Imaging you have multi-sig account and want to login with chat app. In order to signup for this, you need to add an access key to your multi-sig account that points at `chat` contract.

mattlockyer: Makes sense, I will add this. We discussed in Wallet Sync from time to time, so it's a good option to have for apps.

mattlockyer:
RE: adding back cooldown period and default allowance 0, from @evgenykuzyakov on Discord:
> Setting allowance is super dangerous because you may get locked out when using network with super high fees
I'd prefer not to have the allowance, but continue to limit it using rate limiting as you do right now
Key rotation can be done if needed later

Cooldown added back in: https://github.com/near/core-contracts/pull/69/commits/33d4feecc383efc0927fabb61db4eb722d57d29c
Cleaned up with a constant 15 minutes: https://github.com/near/core-contracts/pull/69/commits/7ca2c1507af216abadfe7190ec999793f39127d2

mattlockyer:
@ilblackdragon @evgenykuzyakov
Attempted a fix for AddKey, please take a look. The intention is that the user can add keys for other contracts and provide permissions, but if they try to provide permissions for their own contract, or permissions are missing, the contract will warn and set default permissions respectively.

Also added new request type: `AddFullAccessKey`
The reason: if user of multisig wants to set up a **seed phrase** recovery method, **after** they deploy multisig, using only limited access keys, they should be able to confirm this request and get a seed phrase that is a full access key.
Seed phrase is considered "my keys, my crypto" in our space in general.
We can have special messaging in the wallet(s) warning user what they are doing if they confirm this request.

evgenykuzyakov:
I suggest to split add keys into 2 actions:
- `AddAccessKey` (that’s for access keys)
- `AddSignatureKey` (for internal multi-sig keys)

`AddAccessKey` - is the action that let you add both `FullPermission` and `FunctionCallPermission` access keys.
While `AddSignatureKey` can only be added to itself (receiver == current_account_id). It creates a new multi-sig access key with function call permission and predefined list of method names.

I don't think we need to split `DeleteKey` unless we decide to track access keys internally.

mattlockyer:
@ilblackdragon @vgrichina see above. Motivation for changing add keys is to disambiguate different types of keys that need to be added via requests.
Summary, we need to support the following types of keys:

1. self -> all multisig methods
2. self -> confirm only
3. self -> full access (for seed phrase recovery or 2FA -> Ledger upgrade)
4. other -> function call key with permission & allowance (other contracts)
5. other -> full access key for another account (sub accounts)

@evgenykuzyakov in your proposal, how would we differentiate 1 & 2?

evgenykuzyakov:
Oh, I didn't know you need so much control.

I suggest we only keep `AddKey` without permission checks in the contract. The request has to manually specify permissions for the given key: either full or function call. And then if it's a function call, specify the allowance and method names.

So you can remove logic around `receiver_id == current_id`, and it will be fine.

multisig/src/lib.rs
```diff
@@ -74,110 +124,191 @@ impl MultiSigContract {
...
+    /// Add request for multisig and confirm with the pk that added.
+    pub fn add_request_and_confirm(&mut self, request: MultiSigRequest) -> RequestId {
+        let request_id = self.add_request(request);
+        self.confirm(request_id);
+        request_id
+    }
+
     /// Remove given request and associated confirmations.
     pub fn delete_request(&mut self, request_id: RequestId) {
+        self.valid_request(request_id);
```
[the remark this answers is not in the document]
evgenykuzyakov: Ah. It's already checked in `valid_request`. Please rename it to `assert_valid_request`

multisig/src/lib.rs
```diff
@@ -74,110 +124,191 @@ impl MultiSigContract {
...
 
+    /********************************
+    View methods
+    ********************************/
     pub fn get_request(&self, request_id: RequestId) -> MultiSigRequest {
-        self.requests.get(&request_id).expect("No such request")
+        (self.requests.get(&request_id).expect("No such request")).request
+    }
+
+    pub fn get_num_requests_pk(&self, public_key: PublicKey) -> u32 {
```
evgenykuzyakov, line 310: You may want to change it to `Base58PublicKey`

multisig/README.md
```diff
@@ -46,14 +85,20 @@ pub fn delete_request(&mut self, request_id: RequestId) {
 pub fn confirm(&mut self, request_id: RequestId) -> PromiseOrValue<bool> {
```
evgenykuzyakov, line 85: Missing `add_request_and_confirm` above

multisig/README.md
```diff
@@ -19,18 +19,57 @@ All operations going forward will require `K` signatures to be performed.
...
+    /// Stake with this account.
+    Stake {
+        public_key: Base58PublicKey,
+        stake: U128,
+    },
+    /// Adds key, either new key for multisig or full access key to another account.
+    AddKey {
+        public_key: Base58PublicKey,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        permission: Option<FunctionCallPermission>,
```
evgenykuzyakov, line 43: Is `FunctionCallPermission` explained in the docs?

multisig/README.md
evgenykuzyakov: Also add view methods to README

multisig/src/lib.rs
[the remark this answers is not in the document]
mattlockyer: Done and thanks that makes more sense!

Issue #31, opened 2020-05-22
Refactor lockup to use master account


ilblackdragon:
The suggestion is to change lockup to have 0 access keys.

Instead have a master account that is specified on deployment.

- Lockup contract is only aware of it’s master account
- master account can have a multisig instead of full access key to improve security and allow key management
- master account can authorize staking UI for using lockup contract on it’s behalf of it with `stake,deposit,withdraw,unstake` authorized functions.

Meaning that all public methods inside lockup for user just have a guard of `assert_eq!(env::predessesor(), self.owner_id);` and methods for foundation have `assert_eq!(env::predessesor(), self.foundation_id);`

On the wallets / custody side, this would mean that they can just control master account. A separate page for lockup operations can be created, where users can "Login with NEAR" and then `withdraw` vested tokens. Staking of locked tokens can be done from this page as well (or separate staking app can be used, but needs to be aware of this dynamics).

We can create master accounts via linkdrops with multisig contract as default and only enough $NEAR to cover the contract storage + 3 keys.

CC @kcole16

PR #70, opened 2020-06-18
[0.2.0] Lockup: Replace access keys with external account

evgenykuzyakov:
### `0.2.0`

- Replaced owner's access keys with the owner's account. The access is now controlled through the predecessor account ID similar to NEAR foundation access.
  This allows to be more flexible with the account access including multi-sig implementation.
- The lockup contract account should not have any access keys until the account is fully vested and unlocked.
  Only then the owner can add the full access key.
- Removed methods for adding and removing staking/main access keys.
- Added a view method to get the account ID of the owner.

Fixes: #31


lockup/src/internal.rs
```diff
@@ -97,4 +97,12 @@ impl LockupContract {
             env::panic(b"No NEAR Foundation account is specified in the contract");
         }
     }
+
+    pub fn assert_owner(&self) {
```
fckt, line 101: Maybe it makes sense to have `assert_foundation`, by analogy.

willemneal: There `assert_called_by_foundation` above.

evgenykuzyakov:
@fckt I don't think we can enforce that the owner account exists or have access. Checking the validity of the owner account should be enough.

evgenykuzyakov:
> Do you think there should be a test for a non-owner account calling an owner method?

There is a unit test for this.

fckt:
I mostly mean that we need a deployment script to make sure that deployer (at least) didn't made a mistake

willemneal:
> I mostly mean that we need a deployment script to make sure that deployer (at least) didn't make a mistake on initialization.

Yeah that would be good, but I think that is covered in the main README in the repo.

fckt:
Ok, it's not a blocker. I just think it's it's important to have a "deployer" script which takes all the necessary parameters and then creates accounts, deploys contracts and does asserts.

evgenykuzyakov:
> Ok, it's not a blocker. I just think it's it's important to have a "deployer" script which takes all the necessary parameters and then creates accounts, deploys contracts and does asserts.

Makes sense. I planned to work on this, so we should include it into this repo

PR #72, opened 2020-06-23
[0.2.0] Staking pool: Add pause/resume

### LINKED ISSUE #41: Staking pool contract should have `pause` method for owner
When we do hard-forks, we ask validators who cannot be online during the hard fork to unstake, but there is no way for the owner to unstake all the stake.

It is also useful for maintenance purposes.

/cc @jimmy3dita @nearmax

evgenykuzyakov:
# Changelog

## 0.2.0

- Added new owners methods: `pause_staking` and `resume_staking`. Allows pool owner to unstake everything from the pool for node maintenance.
- Added a new view method `is_staking_paused` to check whether the pool has paused staking.

Fixes #41 

# Test plan

- Added an integration test


staking-pool/src/lib.rs
```diff
@@ -378,12 +385,19 @@ impl StakingContract {
 
     /// Restakes the current `total_staked_balance` again.
     fn restake(&mut self) {
+        if self.paused {
+            return;
```
bowenwang1996, line 389: should we return something that indicates the contract has been paused? Otherwise if people call `stake`, they will think that they have successfully staked and may be confused if they find out that they didn't actually stake.

evgenykuzyakov: Well, they are staked successfully. And the pool will return to normal, once the staking is resumed. People can verify the performance of the pool in the past before staking.

staking-pool/src/lib.rs
```diff
@@ -489,10 +496,41 @@ impl StakingContract {
...
+        self.paused = false;
+        self.restake();
+    }
+
     /********************/
     /* Internal methods */
     /********************/
 
+    /// Asserts that the method was called by the owner.
+    fn assert_owner(&self) {
```
willemneal, line 526: This should become a macro.

evgenykuzyakov: Commented in the issue

PR #74, opened 2020-06-25
Adding core scripts

evgenykuzyakov:
# Setup scripts

## Rebuild all contracts

```bash
./build_all.sh
```

### Set master account

```bash
export MASTER_ACCOUNT_ID=near
```

### Set network environment

```bash
export NODE_ENV=testnet
```

### Deploy

This will deploy the following contracts:

- Voting contract at `vote.<MASTER_ACCOUNT_ID>`
- Whitelist contract at `whitelist.<MASTER_ACCOUNT_ID>`
- Staking pool factory contract at `pool.<MASTER_ACCOUNT_ID>`

It will whitelist the staking pool factory account.

It requires `80` NEAR tokens.

```bash
./deploy_core.sh
```

## Notes

For rebuilding contracts, make sure you have `rust` with `wasm32` target installed.

For deploying, you need to have `near-shell` installed and be logged in with the master account ID.


scripts/deploy/deploy_staking_pool_factory.sh
```diff
@@ -0,0 +1,42 @@
...
+        nearAPI.transactions.transfer("50000000000000000000000000"),
+        nearAPI.transactions.deployContract(fs.readFileSync("../../staking-pool-factory/res/staking_pool_factory.wasm")),
+        nearAPI.transactions.functionCall("new", Buffer.from(JSON.stringify(newArgs)), 10000000000000, "0"),
+    ]);
+END
+)
+
+echo $REPL | near repl
+
+echo "Whetelisting staking pool factory $ACCOUNT_ID on whitelist contract $WHITELIST_ACCOUNT_ID"
```
bowenwang1996, line 28:
```suggestion
echo "Whitelisting staking pool factory $ACCOUNT_ID on whitelist contract $WHITELIST_ACCOUNT_ID"
```

scripts/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Set master account
+
+```bash
+export MASTER_ACCOUNT_ID=near
+```
+
+### Set network environment
+
+```bash
+export NODE_ENV=testnet
```
bowenwang1996, line 20: is this actually a legitimate environment? I think the environment for testnet is `development`

fckt:
Looks ok to me. I was surprised to see bash scripts, but not just JS scripts 🧐.. 

Should we add these scripts into CI to check that scripts are operational (in case of possible changes)?

scripts/README.md
evgenykuzyakov: Yes, it deploys to `testnet`

scripts/build_all.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -ex
+
+for p in lockup multisig staking-pool staking-pool-factory voting whitelist
+do
+  pushd ../${p}
```
willemneal, line 6:
```suggestion
 (cd ${p} && ./build.sh)
```
This will use a subprocess and avoid changing the parent script's pwd.

scripts/build_all.sh
```diff
@@ -0,0 +1,9 @@
...
+  ./build.sh
```
willemneal, line 7:
```suggestion
```

scripts/build_all.sh
```diff
@@ -0,0 +1,9 @@
...
+  popd
```
willemneal, line 8:
```suggestion
```

willemneal:
It looks good to me too, though I would prefer if it was `echo ".load file" | near repl`.

scripts/tests/test_voting.sh
```diff
@@ -0,0 +1,119 @@
...
+set -e
+
+# nearup localnet --num-nodes 5 --docker-image "nearprotocol/nearcore:master" --overwrite
+nearup localnet --num-nodes 5 --binary-path /Users/ekwork/code/nearcore/target/debug/ --overwrite
+
+export MASTER_ACCOUNT_ID=node0
+export NODE_ENV=local
+
+stop_nodes() {
+  echo "STOOOP THE NODES!"
```
willemneal, line 11: Lol should be `STOOOP THE NOOODES`

chain-police: 😀

scripts/deploy_lockup.sh
```diff
@@ -0,0 +1,121 @@
...
+
+if [ "$PREFIX_RE" = "OK" ]; then
+  ACCOUNT_ID="$ACCOUNT_PREFIX.${MASTER_ACCOUNT_ID}"
+else
+  echo "Invalid new account prefix."
+  exit 1
+fi
+
+
+LOCKUP_ACCOUNT_ID="lockup.$ACCOUNT_ID"
```
ilblackdragon, line 36: we are planning to do instead `lockup` account that we control, and deploy under `$ACCOUNT_ID.$LOCKUP_ID` because otherwise we can't actually deploy this undert this account as we don't have perm

evgenykuzyakov: make sense

scripts/deploy_lockup.sh
```diff
@@ -0,0 +1,121 @@
...
+    echo "The minimum balance has to be $MINIMUM_BALANCE. Try again."
+  fi
+done
+
+
+
+#
+#if
+#
+#echo "Deploying staking pool factory contract to $ACCOUNT_ID with 50 NEAR"
```
bowenwang1996, line 86: what happened here?

PR #76, opened 2020-07-01
add example to staking pool factory readme

bowenwang1996:
Add example usage of the staking pool factory.

### DIFF
--- a/staking-pool-factory/README.md
+++ b/staking-pool-factory/README.md
@@ -61,3 +61,18 @@ pub fn on_staking_pool_create(
     predecessor_account_id: AccountId,
 ) -> PromiseOrValue<bool>;
 ```
+
+### Usage
+
+Contract initialization:
+```bash
+near call <staking_pool_factory> new '{"staking_pool_whitelist_account_id": <whitelist_contract>}'
+```
+where `<whitelist_contract>` it the account id of the whitelist contract that this factory
+is associated with.
+
+Staking pool creation:
+```bash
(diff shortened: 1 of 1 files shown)


staking-pool-factory/README.md
```diff
@@ -61,3 +61,18 @@ pub fn on_staking_pool_create(
...
 ```
+
+### Usage
+
+Contract initialization:
+```bash
+near call <staking_pool_factory> new '{"staking_pool_whitelist_account_id": <whitelist_contract>}'
+```
+where `<whitelist_contract>` it the account id of the whitelist contract that this factory
+is associated with.
```
evgenykuzyakov, line 72: I think this should be inlined with the deployment, but it's okay to have it as an initialization example with the note of caution.

staking-pool-factory/README.md
```diff
@@ -61,3 +61,18 @@ pub fn on_staking_pool_create(
...
+
+Staking pool creation:
+```bash
+near call <staking_pool_factory> create_staking_pool '{"staking_pool_id": <staking_pool_id>, "owner_id": <owner_id>, "stake_public_key": <stake_public_key>, "reward_fee_fraction": <reward_fraction>}' 
+```
+Here `reward_fee_fraction` is an object like `{"numerator": 10, "denominator": 100}`.
```
evgenykuzyakov, line 78: Please explain every parameter, because people are still confused. Or point to the `staking-pool` contract

PR #79, opened 2020-07-15
[0.2.0] voting: Replace inner Map with HashMap

evgenykuzyakov:
Update voting implementation to use `HashMap` instead of persistent `UnorderedMap`. It might be too expensive to iterate through each element of the persistent `UnorderedMap` because it issues at least 2 read calls for each element.

The base gas cost of a read is quite expensive to do this for 100 validators.

The good guidance for using `HashMap` instead of `UnorderedMap` should be whether you need to iterate on the full map. If you ever need to iterate on all elements of the `UnorderedMap`, then you probably should use `HashMap` or a simple `Vec`.

This change also bumps `near_sdk` dependency for the voting contract and exposes 2 new view methods: `get_votes` and `get_total_voted_stake`


voting/src/lib.rs
```diff
@@ -47,13 +47,13 @@ impl VotingContract {
         assert!(self.result.is_none(), "Voting has already ended");
         let cur_epoch_height = env::epoch_height();
         if cur_epoch_height != self.last_epoch_height {
-            for account_id in self.votes.keys().into_iter().collect::<Vec<_>>() {
+            let votes = std::mem::take(&mut self.votes);
+            for (account_id, account_voted_stake) in votes.into_iter() {
```
bowenwang1996, line 51: `into_iter` is unnecessary

evgenykuzyakov: Good point. It produced the same binary.

PR #81, opened 2020-07-17
Voting contract test with localnet

evgenykuzyakov:


scripts/tests/test_voting.sh
```diff
@@ -0,0 +1,68 @@
...
+# Deploying core accounts
+(cd .. && ./deploy_core.sh)
+
+for (( i=1; i<=$LAST_NODE; i++ )); do
+  ACCOUNT_ID="node${i}"
+  near stake $ACCOUNT_ID "ed25519:7PGseFbWxvYVgZ89K1uTJKYoKetWs7BJtbyXDzfbAcqX" 0
+done;
+
+echo "Sleeping 3 minutes (for 3 epochs)"
+sleep 181
```
bowenwang1996, line 28: this might be too tight.

scripts/tests/test_voting.sh
```diff
@@ -0,0 +1,68 @@
...
+
+echo "The only current validator should be the node0"
+near validators current
+
+for (( i=1; i<=$LAST_NODE; i++ )); do
+  ACCOUNT_ID="node${i}"
+  near deploy --wasmFile="../../staking-pool/res/staking_pool.wasm" --accountId=$ACCOUNT_ID
+  PUBLIC_KEY=$(grep -oE 'ed25519:[^"]+' ~/.near/localnet/node$i/validator_key.json | head -1)
+  near call $ACCOUNT_ID new "{\"owner_id\": \"$OWNER_ACCOUNT_ID\", \"stake_public_key\": \"$PUBLIC_KEY\", \"reward_fee_fraction\": {\"numerator\": 10, \"denominator\": 100}}" --accountId=$OWNER_ACCOUNT_ID
```
bowenwang1996, line 37: Probably better to use single quotes on the outside to make it more readable

evgenykuzyakov: It looks like `$OWNER_ACCOUNT_ID` is not replaced with the variable value within single quotes.

PR #82, opened 2020-07-17
Allow to keep names in contracts.

olonho:


build_all.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+for d in */ ; do
```
evgenykuzyakov, line 4: Contracts has to be built in the specific order. Especially, the factory contract has to be built after staking pool, because it hardcodes the code inside.

evgenykuzyakov (approved):

ReClicker (approved):

ReClicker (approved):

PR #84, opened 2020-07-28
fix: DeleteKey action should clean up storage for public_key

mattlockyer:
@evgenykuzyakov/@ilblackdragon let me know if this works or you want a different approach.

Problem: deleting a `public_key` can leave "orphan" requests that were not confirmed and we also track `num_requests_pk`.

Solution: clean this up when the key is deleted e.g. during a key rotation or de-authorization from the wallet.


multisig/src/lib.rs
```diff
@@ -201,7 +201,18 @@ impl MultiSigContract {
                     }
                 }
                 MultiSigRequestAction::DeleteKey { public_key } => {
-                    promise.delete_key(public_key.into())
+                    let pk:PublicKey = public_key.into();
+                    // delete outstanding requests by public_key
```
evgenykuzyakov, line 205: This action can be used on other `receiver_id`. Also the key might not be part of the multi-key. Just need to be careful there

multisig/src/lib.rs
```diff
@@ -201,7 +201,18 @@ impl MultiSigContract {
...
+                    let request_ids = self.list_request_ids();
+                    for request_id in request_ids {
+                        let request_with_signer = self.requests.get(&request_id).expect("No such request");
```
evgenykuzyakov, line 208: You can iterate on values instead and remember indices that has to be removed.
Then you can collect keys by index (request Ids) and only then remove such requests in a loop.

This will be more efficient because you will skip a few reads required to first get the request_id for items that don't match the signer_pk.

For this you can use `keys_as_vector` and `values_as_vector` methods of `UnorderedMap`

multisig/src/lib.rs
mattlockyer: > This action can be used on other `receiver_id`. Also the key might not be part of the multi-key. Just need to be careful there

Wondering if we should implement: https://github.com/near/core-contracts/issues/86 ?

multisig/src/lib.rs
```diff
@@ -192,16 +185,29 @@ impl MultiSigContract {
...
+                    let pk: PublicKey = public_key.into();
+                    // delete outstanding requests by public_key
+                    let request_ids: Vec<u32> = self
+                        .requests
+                        .iter()
+                        .filter(|(_k, r)| r.signer_pk == pk)
+                        .map(|(k, _r)| k)
+                        .collect();
+                    for request_id in request_ids {
+                        self.remove_request(request_id);
```
evgenykuzyakov, line 206: It's somewhat inefficient to call `remove_request` because it updates `num_requests_pk`, and this requires ton of read/writes. I think it's better to copy only the relevant part there, because you'll nuke the `num_requests_pk` for this entry at the end anyway

PR #87, opened 2020-07-30
[0.4.0] staking pool: Add helper methods for delegators

evgenykuzyakov:
- Internal refactoring. Moving internal methods to `internal.rs`
- Added 4 new delegator methods:
    - `deposit_and_stake` - to deposit and stake attached balance in one call.
    - `stake_all` - to stake all unstaked balance.
    - `unstake_all` - to unstake all staked balance.
    - `withdraw_all` - to withdraw all unstaked balance.

Release `0.4.0` of the staking pool
And `0.1.4` for staking pool factory.


staking-pool/src/lib.rs
```diff
@@ -215,207 +217,100 @@ impl StakingContract {
     pub fn deposit(&mut self) {
         let need_to_restake = self.internal_ping();
 
+        let _amount = self.internal_deposit();
```
bowenwang1996, line 220:
```suggestion
        self.internal_deposit();
```

staking-pool/src/lib.rs
```diff
@@ -215,207 +217,100 @@ impl StakingContract {
...
+
+        if need_to_restake {
+            self.internal_restake();
+        }
+    }
+
+    /// Deposits the attached amount into the inner account of the predecessor and stakes it.
+    #[payable]
+    pub fn deposit_and_stake(&mut self) {
+        let _need_to_restake = self.internal_ping();
```
bowenwang1996, line 230:
```suggestion
        self.internal_ping();
```

jimmy3dita:
Which are the BetaNet and TestNet "pool factories" that users have to call? I will update the guides with this new release.
Also, has an impact on the staking cheatsheet, will have to update that as well

PR #88, opened 2020-08-10
Fast fun token

evgenykuzyakov:
I've implemented a simple fungible token contract in pure Rust wasm that allows to transfer fungible tokens using `U256` balance.  Probably can be simplified to use `u128` instead to make the size even smaller.

But I chose U256 to match EVM like math.

It's `1288` bytes total in Wasm.


ilblackdragon:
Should we keep this here or should it go somewhere else?

evgenykuzyakov:
Don't need it. Let's close

PR #90, opened 2020-08-17
Include code hash to last commit mapping with stats

vidorge:


vidorge:
@evgenykuzyakov We wanted to list all staking contract hashes so we know which borsh schema to use for deserialization of the contract state. Turns out that the compilation process isn't deterministic and provides different outputs on different platforms so this idea might not be the way to go.

Stefano Pepe asked if we could put together a compatibility table and do a PR, but looks like interest has died down.

The reason we are decoding contract state instead of using the contract view methods is because it is much more performant than separately invoking methods for a large number of contracts.

I will be closing this.

PR #95, opened 2020-08-26
Add Minify Script

mattlockyer:
Adding minify script to multisig contract /res in order to minify for releases.


vgrichina:
@nearmax @evgenykuzyakov  @ilblackdragon 

I'm a bit confused about `wasm-opt` stuff. Doesn't it raise exactly same concerns as AssemblyScript usage? If we going to use binaryen on Rust binaries, what is the reason to believe there is any less exposure to binary generation bugs than with AssemblyScript? Binaryen is even more dependent on single contributor maintaining it.

mattlockyer:
Agree. Usage was premature. Closing this PR.

nearmax:
minify script corrupted our contracts in the past. Also, @vgrichina is right.

PR #96, opened 2020-08-27
Lockup: Make release_duration compatible with vesting schedule

evgenykuzyakov:
### `1.0.0`

- Make `release_duration` independent from the `vesting_schedule`. They are not allowed to be used simultaneously.
- Internal. Remove some JSON serialization on inner structures.

## Tets plan:
- Added unit tests


lockup/src/getters.rs
```diff
@@ -61,24 +61,47 @@ impl LockupContract {
...
-                self.lockup_information
-                    .lockup_timestamp
-                    .map_or(0, |t| t.0),
+                    .saturating_add(self.lockup_information.lockup_duration),
+                self.lockup_information.lockup_timestamp.unwrap_or(0),
             );
-            if lockup_timestamp <= env::block_timestamp() {
-                return self.get_unvested_amount();
+            let block_timestamp = env::block_timestamp();
+            if lockup_timestamp <= block_timestamp {
```
bowenwang1996, line 76: where is the lockup cliff?

evgenykuzyakov: It's `lockup_duration` above.

PR #100, opened 2020-09-07
fix: edge case

### LINKED ISSUE #99: edge case bug - contract account with low balance
If the staking pool is created on an account with balance = 1_000_000_000_000, then total_stake_shares is 0, and several view functions start to panic

**Note:** I'm preparing a PR to handle this edge case

To reproduce, add this unit test: (the 4th parameter is the account initial balance)

```
    #[test]
    fn test_low_staked_balance() {
        let mut emulator = Emulator::new(
            owner(),
            "KuTCtARNzxZQ3YvXDeLjx83FDqxv2SdQTSbiq876zR7".to_string(),
            zero_fee(),
            STAKE_SHARE_PRICE_GUARANTEE_FUND, //account initial balance, minimum, edge case
        );
        //bob deposits 1M
        let deposit_amount = ntoy(1_000_000);
        emulator.update_context(bob(), deposit_amount);
        emulator.contract.deposit();
        emulator.amount += deposit_amount;

        //bob wants to know how much unstaked balance he has
        emulator.update_context(bob(), 0);
        assert_eq!(
            emulator.contract.get_account_unstaked_balance(bob()).0,
            deposit_amount
        );
        //panic!
        //thread 'tests::test_low_staked_balance' panicked at 'The total number of stake shares can't be 0', src/internal.rs:297:9

    }

```

luciotato:
fixes #99

### DIFF
--- a/staking-pool/src/lib.rs
+++ b/staking-pool/src/lib.rs
@@ -183,6 +183,10 @@ impl StakingContract {
             "The owner account ID is invalid"
         );
         let account_balance = env::account_balance();
+        assert!(
+            account_balance > STAKE_SHARE_PRICE_GUARANTEE_FUND,
+            "Account balance too low"
+        );
         let total_staked_balance = account_balance - STAKE_SHARE_PRICE_GUARANTEE_FUND;
         assert_eq!(
             env::account_locked_balance(),
@@ -514,10 +518,11 @@ mod tests {
             owner: String,
             stake_public_key: String,
             reward_fee_fraction: RewardFeeFraction,
+            initial_balance: Balance,
         ) -> Self {
             let context = VMContextBuilder::new()
                 .current_account_id(owner.clone())
-                .account_balance(ntoy(30))
+                .account_balance(initial_balance)
                 .finish();
             testing_env!(context.clone());
             let contract = StakingContract::new(
@@ -530,7 +535,7 @@ mod tests {
             Emulator {
                 contract,
                 epoch_height: 0,
-                amount: ntoy(30),
+                amount: initial_balance,
                 locked_amount: 0,
                 last_total_staked_balance,
                 last_total_stake_shares,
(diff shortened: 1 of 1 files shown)


staking-pool/src/lib.rs
```diff
@@ -514,10 +518,11 @@ mod tests {
             owner: String,
             stake_public_key: String,
             reward_fee_fraction: RewardFeeFraction,
+            initial_balance: Balance,
```
luciotato, line 521: add a 4th parameter for testing, initial balance

staking-pool/src/lib.rs
```diff
@@ -183,6 +183,10 @@ impl StakingContract {
             "The owner account ID is invalid"
         );
         let account_balance = env::account_balance();
+        assert!(
```
luciotato, line 186: avoid `total_staked_balance`=0

evgenykuzyakov: Realistically, this is impossible, because the storage required for the contract itself covers this.

staking-pool/src/lib.rs
```diff
@@ -183,6 +183,10 @@ impl StakingContract {
...
+            account_balance > STAKE_SHARE_PRICE_GUARANTEE_FUND,
```
evgenykuzyakov, line 187: Why this is strictly higher?

staking-pool/src/lib.rs
luciotato: I see. The contract will stop working way before that.
Question: Who pays the storage fees? It comes from the staked NEARs?

staking-pool/src/lib.rs
luciotato: because if `account_balance == STAKE_SHARE_PRICE_GUARANTEE_FUND` then in the next line total_staked_balance becomes zero. `total_staked_balance` is used in several share calculations as denominator, so `total_staked_balance==0` leads to div by zero panics. (i.e. `fn num_shares_from_staked_amount_rounded_down/up`)

PR #106, opened 2020-09-15
Adding hash for vesting schedule

ilblackdragon:
By @evgenykuzyakov: 
- recompiled the contract with docker
- all integration and unit tests pass
- Updated README.
- Added a view method to get lockup `VestingInformation` to build ✌️DeFi✌️.

### Private vesting schedule

Since vesting schedule usually starts at the date of employment it allows to de-anonymize the owner of the lockup contract.
To keep the identity private, the contract allows to hash the vesting schedule with some random salt and keep store the
hash instead of the raw vesting schedule information. In case the NEAR foundation has to terminate the vesting schedule,
it will provide the raw vesting schedule and the salt, effectively revealing the vesting schedule. The contract then
will compare the hash with the internal hash and if they match proceed with the termination.

NOTE: The private vesting schedule can only be used, if the lockup release period and the lockup duration are effectively
shadowing the vesting duration. Meaning that the lockup release ends later than the vesting release and the lockup duration
ends after the vesting cliff.
Once the lockup schedule starts before the vesting schedule (e.g. employment starts after the transfers are enabled),
the vesting schedule can't be kept private anymore.

### `2.0.0`

- Changed `vesting_schedule` initialization argument to allow it to hide the vesting schedule behind a hash to keep it private.
- Added view method `get_vesting_information` to view internal vesting information.


lockup/src/getters.rs
```diff
@@ -92,33 +94,36 @@ impl LockupContract {
                         0
                     };
 
-                return (std::cmp::max(
+                let unvested_amount = match &self.vesting_information {
+                    VestingInformation::VestingSchedule(vs) => self.get_unvested_amount(vs.clone()),
+                    VestingInformation::Terminating(terminating) => terminating.unvested_amount,
+                    // There is either no information because vesting is private.
```
evgenykuzyakov, line 100:
```suggestion
                    // Vesting is private, so we can assume the vesting started before lockup date.
```

lockup/src/types.rs
```diff
@@ -115,13 +118,23 @@ impl VestingSchedule {
     }
 }
 
+#[derive(Serialize, Deserialize)]
+#[serde(crate = "near_sdk::serde")]
+#[serde(tag = "type", content = "value")]
+pub enum VestingScheduleOrHash {
+    Hash(String),
```
evgenykuzyakov, line 125: Use `Base64VecU8`
```suggestion
    Hash(Base64VecU8),
```

lockup/src/foundation.rs
```diff
@@ -8,10 +8,10 @@ impl LockupContract {
     /// Requires 25 TGas (1 * BASE_GAS)
     ///
     /// Terminates vesting schedule and locks the remaining unvested amount.
-    pub fn terminate_vesting(&mut self) {
+    pub fn terminate_vesting(&mut self, vesting_schedule: VestingSchedule, salt: Vec<u8>) {
```
evgenykuzyakov, line 11: For salt use `Base64VecU8`

lockup/src/lib.rs
```diff
@@ -145,7 +140,7 @@ pub struct LockupContract {
     pub staking_information: Option<StakingInformation>,
 
     /// The account ID that the NEAR Foundation, that has the ability to terminate vesting.
-    pub foundation_account_id: Option<AccountId>,
+    pub foundation_account_id: AccountId,
```
evgenykuzyakov, line 143: It's cleaner if no vesting requires that there is no `foundation_account_id`

lockup/src/lib.rs
```diff
@@ -168,10 +163,10 @@ impl LockupContract {
     /// - `transfers_information` - the information about the transfers. Either transfers are
     ///    already enabled, then it contains the timestamp when they were enabled. Or the transfers
     ///    are currently disabled and it contains the account ID of the transfer poll contract.
-    /// - `vesting_schedule` - if present, describes the vesting schedule for employees. Vesting
-    ///    schedule affects the amount of tokens the NEAR Foundation will get in case of
+    /// - `vesting_schedule` - Hash of vesting schedule with salt.
```
evgenykuzyakov, line 166: Need to update the comment

evgenykuzyakov (changes requested):

lockup/src/types.rs
```diff
@@ -115,13 +118,23 @@ impl VestingSchedule {
     }
 }
 
+#[derive(Serialize, Deserialize)]
```
evgenykuzyakov, line 121: Only need `Serialize`

lockup/src/lib.rs
```diff
@@ -199,12 +198,6 @@ impl LockupContract {
             env::is_valid_account_id(staking_pool_whitelist_account_id.as_bytes()),
             "The staking pool whitelist account ID is invalid"
         );
-        if foundation_account_id.is_some() {
-            assert!(
-                vesting_schedule.is_some(),
-                "Foundation account can't be added without vesting schedule"
-            )
-        }
```
evgenykuzyakov, line 207: This check is still needed if foundation is provided, but vesting schedule is none

lockup/src/lib.rs
```diff
@@ -850,22 +874,6 @@ mod tests {
...
-    }
-
-    #[test]
-    #[should_panic(expected = "Foundation account can't be added without vesting schedule")]
-    fn test_init_foundation_key_no_vesting_with_release() {
-        let context = basic_context();
-        testing_env!(context.clone());
-        new_contract(true, None, Some(to_nanos(YEAR).into()), true);
-    }
-
```
evgenykuzyakov, line 868: Keep those

lockup/src/lib.rs
```diff
@@ -926,53 +934,65 @@ mod tests {
...
+            to_nanos(YEAR).into(),
+            None,
+            TransfersInformation::TransfersEnabled {
+                transfers_timestamp: to_ts(GENESIS_TIME_IN_DAYS).into(),
+            },
+            VestingScheduleOrHash::VestingSchedule(vesting_schedule.clone()),
+            None,
+            AccountId::from("whitelist"),
+            Some(account_foundation()),
+        );
```
evgenykuzyakov, line 958: Can you update `new_contract` with whatever changes needed to avoid initializing the contract here?

lockup/src/types.rs
```diff
@@ -45,7 +48,7 @@ pub struct LockupInformation {
 }
 
 /// Contains information about the transfers. Whether transfers are enabled or disabled.
-#[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize)]
+#[derive(BorshDeserialize, BorshSerialize, Deserialize, Serialize, Debug)]
```
evgenykuzyakov, line 51: It seems `Debug` increases the size of the contract a lot. See +57Kb

evgenykuzyakov: Apparently `Debug` is not the largest offender here. It's serde with tag, in particular this line:
```rust
#[serde(tag = "type", content = "value")]
```

It adds around 30Kb

evgenykuzyakov:
Assigned to @ilblackdragon for review as well.

lockup/src/lib.rs
evgenykuzyakov: It looks like I pasted them back above

lockup/src/types.rs
```diff
@@ -165,3 +184,13 @@ pub struct TerminationInformation {
 /// The result of the transfer poll.
 /// Contains The timestamp when the proposal was voted in.
 pub type PollResult = Option<WrappedTimestamp>;
+
+pub fn hash_vesting_schedule(vesting_schedule: &VestingSchedule, salt: &[u8]) -> Hash {
```
nearmax, line 188: I would suggest defining `VestingScheduleWithSalt` structure and serialize it as borsh.

lockup/src/internal.rs
```diff
@@ -30,12 +30,19 @@ impl LockupContract {
...
-            env::panic(b"There is no vesting in progress");
-        }
+    pub fn assert_vesting(&self, vesting_schedule: &VestingSchedule, salt: &[u8]) {
+        match &self.vesting_information {
+            VestingInformation::VestingHash(hash) => assert_eq!(
+                &hash_vesting_schedule(vesting_schedule, salt),
+                &hash.0,
+                "Presented vesting doesn't match or vesting was terminated"
+            ),
+            VestingInformation::VestingSchedule(vs) => assert_eq!(
```
nearmax, line 40: It seems like we support a scenario when vesting schedule is saved directly in the contract but then foundation still needs to provide it to call contract methods. It is just a weird UX. We could keep it to make things simple, but we should explain it in the comment, otherwise people would think there is a deeper meaning to this.

lockup/src/types.rs
nearmax: FYI @mikedotexe @chadoh ^

lockup/src/lib.rs
```diff
@@ -222,12 +212,23 @@ impl LockupContract {
...
+                    "Foundation account can't be added without vesting schedule"
+                );
+                VestingInformation::None
+            }
+            Some(VestingScheduleOrHash::VestingHash(hash)) => VestingInformation::VestingHash(hash),
+            Some(VestingScheduleOrHash::VestingSchedule(vs)) => {
+                VestingInformation::VestingSchedule(vs)
+            }
         };
+        assert!(
```
nearmax, line 228: This check looks irrelevant to adding vesting hash. Did we miss a bug in the past?

lockup/src/internal.rs
evgenykuzyakov: Yep. Good point. Will update `terminate_vesting` method to take optional `VestingScheduleWithSalt`.

lockup/src/lib.rs
evgenykuzyakov: Not really. Originally, I wrote the vesting schedule to be used for linear release without realizing we need both linear release and vesting schedule. So it was possible to have vesting schedule without foundation account in the past.

lockup/src/types.rs
evgenykuzyakov: done

bowenwang1996 (approved):

evgenykuzyakov (commented):
I'm going to approve it. @ilblackdragon once you review it, feel free to merge

evgenykuzyakov (approved):

vgrichina (approved):

PR #109, opened 2020-09-28
Removed deposit_and_stake, seems deprecated

mattlockyer:
Cannot find in source anywhere, tripped me up... `deposit_and_stake`


evgenykuzyakov:
It's here: https://github.com/near/core-contracts/blob/215d4ed2edb563c47edd961555106b74275c4274/lockup/src/owner.rs#L127

PR #117, opened 2020-11-11
Epoch teller

evgenykuzyakov:
# Simplest contract to tell current epoch height

Total size `278` bytes

One method to return the current epoch height from `100` to `999`.
Deployed on `epoch.berryclub.ek.near`

To view the current epoch on mainnet call:
```bash
near view epoch.berryclub.ek.near e
```


epoch-teller/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn epoch_height() -> u64;
+    fn value_return(value_len: u64, value_ptr: u64);
+}
+
+#[no_mangle]
+pub unsafe fn e() {
+    let epoch = epoch_height();
+    let a = [b'"', b'0' + ((epoch / 100) % 10) as u8, b'0' + ((epoch / 10) % 10) as u8,  b'0' + (epoch % 10) as u8, b'"'];
+    value_return(a.len() as _, a.as_ptr() as _);
+}
```
frol, line 27: If we want to use it to serve as an example of any kind, I suggest we don't use `as _`, and instead use a more explicit syntax:

```rust
#![no_std]
#![feature(core_intrinsics)]

use core::convert::TryFrom;

#[panic_handler]
#[no_mangle]
pub fn on_panic(_info: &::core::panic::PanicInfo) -> ! {
    unsafe {
        ::core::intrinsics::abort();
    }
}

extern "C" {
    fn epoch_height() -> u64;
    fn value_return(value_len: u64, value_ptr: u64);
}

#[no_mangle]
pub fn e() {
    let epoch = unsafe { epoch_height() };

    let epoch_height_json_str = [
        b'"',
        b'0' + ((epoch / 100) % 10) as u8,
        b'0' + ((epoch / 10) % 10) as u8,
        b'0' + (epoch % 10) as u8,
        b'"',
    ];

    // NOTE: try_from().expect() are optimized away in release as they are no-op
    let return_value_length: u64 = TryFrom::try_from(epoch_height_json_str.len())
        .expect("usize is expected to be <= u64. It should be unreachable");
    let return_value_pointer: u64 = TryFrom::try_from(epoch_height_json_str.as_ptr() as usize)
        .expect("usize is expecte to be <= u64. It should be unreachable");
    unsafe {
        value_return(return_value_length, return_value_pointer);
    }
}
```

BTW, the Wasm binary size remains the same - 278 bytes.

ilblackdragon:
I'm curious why is this contract needed?

evgenykuzyakov:
Don't need it. Let's close

PR #120, opened 2020-11-20
wNear

vince0656:
- [x] NEP21 Compatible
- [x] Behaves like WETH9
- [x] Has a subset of features from WETH10
- [x] Unit tests in place for core functionality
- [x] Deployed to `wnear.testnet`


nearmax:
@evgenykuzyakov we badly need to figure out how we are going to do contract inheritance to reduce code duplications between different versions of NEP21.

evgenykuzyakov:
> we badly need to figure out how we are going to do contract inheritance to reduce code duplications between different versions of NEP21.

That's doable by reusing a crate. I wonder if we can put `#[near-bindgen]` macro to a crate. Otherwise we can just you a library with helper methods

w-near/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "w_near"
+version = "0.1.0"
+authors = ["Near Inc <hello@nearprotocol.com>"]
```
evgenykuzyakov, line 4: Feel free to change `authors`

w-near/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+cd "`dirname $0`"
+source ./flags.sh
+cargo build --target wasm32-unknown-unknown --release
+cp target/wasm32-unknown-unknown/release/fungible_token.wasm ./res/
```
evgenykuzyakov, line 6:
```suggestion
cp target/wasm32-unknown-unknown/release/w_near.wasm ./res/
```

w-near/src/lib.rs
```diff
@@ -0,0 +1,1351 @@
+/**
```
evgenykuzyakov, line 1: It makes sense to split it into 2 or 3 files to make easier to compare to existing fungible-token implementation

`lib.rs` - main entrypoint
`fungible_token.rs` - NEP21 implementation
`w_near.rs` - extra logic to mint and burn

w-near/src/lib.rs
```diff
@@ -0,0 +1,1351 @@
...
+    /// Deposit NEAR from the predecessor account and send wNear to a specific receiver_id
+    /// Requirements:
+    /// * `receiver_id` cannot be this contract
+    /// * `receiver_id` must be a valid account Id
+    /// * `amount` must be a positive integer
+    /// * Caller of the method has to attach deposit enough to cover:
+    ///   * The `amount` of wNear tokens being minted, and
+    ///   * The storage difference at the fixed storage price defined in the contract.
+    #[payable]
+    pub fn deposit_to(&mut self, receiver_id: AccountId, amount: U128) {
```
evgenykuzyakov, line 110: With `near-sdk = 2.0.0` crate, you can use `ValidAccountId` argument type from `near_sdk::json_types` mod.
It's a wrapped `AccountId` that verifies account ID validity at deserialization time, so you don't need to check `env:: is_valid_account_id` later. 
Then you can cast it to `AccountId` using `into`. 

```suggestion
    pub fn deposit_to(&mut self, receiver_id: ValidAccountId, amount: U128) {
        let receiver_id: AccountId = receiver_id.into();
```

evgenykuzyakov (changes requested):

w-near/src/lib.rs
vince0656: @evgenykuzyakov I have replaced `AccountId` where I could to `ValidAccountId`. However, `deposit_to()` is called by `deposit` and the first argument passed is `env::predecessor_account_id()` which returns a string. As far as I can see, there is no way to cast from `AccountId` to `ValidAccountId`. Therefore, for `deposit_to()` and `withdraw_to()`, I have left the param type as `AccountId` for `receiver_id`

w-near/src/lib.rs
vince0656: @evgenykuzyakov I have now split this up. Granted it is split up into more files than suggested but makes for clean separation between internal methods and public. This should now make further review more straight forward but let me know what you think

w-near/src/lib.rs
vince0656: @evgenykuzyakov In the end I had to revert all the param changes I did back to all being of type `AccountId`. I couldn't find a way of creating a `ValidAccountId` from a string which meant that all my tests were failing. I am happy to use `ValidAccountId` if I can do the conversion from string. Let me know what you think

mfornet (approved):
LGTM

evgenykuzyakov (approved):
Let's merge this and later fix a few things in a separate PR.
- `ValidAccountId` can be constructed from tests using `account_id.try_into().unwrap()`
- Rebuild the wasm binary with the Docker contract builder from near-sdk repo. This allows to build the same reproducible wasm binary on different machines.

evgenykuzyakov:
- [x] NEP21 Compatible
- [x] Behaves like WETH9
- [x] Has a subset of features from WETH10
- [x] Unit tests in place for core functionality
- [x] Deployed to `wnear.testnet`

TODO: 
- `ValidAccountId` can be constructed from tests using `account_id.try_into().unwrap()`
- Rebuild the wasm binary with the Docker contract builder from near-sdk repo. This allows to build the same reproducible wasm binary on different machines.

Issue #86, opened 2020-07-29
[multisig] Maintain internal set of keys with extendable options


ilblackdragon:
Because NEAR runtime doesn't have a way to query access keys that current account has, contract must have an internal set of keys. This would also allow to configure different options around specific keys.

```
/// Contain list of options that multisig supports for given key.
/// For example: no confirmation for some small amounts or whitelisted accounts.
struct KeyOption {
   ...
}

struct Multisig {
  ...
  keys: HashMap<Key, KeyOption>
}

impl Multisig {
  ..
  fn new(keys: Vec<Base58PublicKey>) -> Self {
       ... actually add keys to access keys ...
  }
```

ilblackdragon:
Currently we can easily mess up keys at the setup (for example by not adding all the required allowed methods). This would address this problem.

Also `AddKey` action should be split into `AddKey` for other receivers and `AddMultisigKey` to add new access key for this contract.

PR #125, opened 2021-01-11
[multisig] Add option for multisig to be managed by other accounts

### LINKED ISSUE #124: Add support for external accounts to manage multisig
Currently multisig is only managed by access keys of the account containing multisig.

This is not always convenient, as users might already have existing accounts but want to band together to create multisig for other purposes (managing treasury, upgradability, etc)

Proposal to extend multisig contract to support both PublicKey and AccountId as members of it.

This will also address #86 by keeping internal track of the keys / accounts added and remove and preventing potential issues with removing more members of multisig than number of confirmations.

ilblackdragon:
Extends multisig contract to be managed either by AccessKeys or by other accounts (all together called "members" of multisig)

Fixes #124 
Ref #86


multisig/src/lib.rs
```diff
@@ -100,42 +141,106 @@ impl MultiSigContract {
...
+        assert!(
+            self.members.len() - 1 >= self.num_confirmations as u64,
+            "Removing given member will make total number of members below number of confirmations"
+        );
+        // delete outstanding requests by public_key
+        let member_str: String = member.clone().into();
+        let request_ids: Vec<u32> = self
+            .requests
+            .iter()
+            .filter(|(_k, r)| r.member == member_str)
```
bowenwang1996, line 208: why not just `filter_map`

bowenwang1996 (approved):

bowenwang1996:
@evgenykuzyakov could you review as well?

multisig/src/lib.rs
```diff
@@ -59,32 +63,69 @@ pub enum MultiSigRequestAction {
...
+        match member {
+            MultisigMember::Account(account_id) => account_id,
+            MultisigMember::AccessKey(key) => String::from(&Base58PublicKey(key.to_vec())),
+        }
+    }
+}
+
+impl From<String> for MultisigMember {
+    fn from(s: String) -> Self {
+        if s.contains(":") {
```
evgenykuzyakov, line 100: `:` is not a requirement for the public key string representation right now. This can lead to abuse if the base58 public key was only lowercase alphanumeric.

evgenykuzyakov (dismissed):
We can't push it into the current multi-sig, because it makes multi-sig non-upgradable from the current STATE.

Better to fork it into a separate folder.

multisig/src/lib.rs
ilblackdragon: I was thinking to add this as requirement.
But may be something like [{"account_id": "xyz"}, {"public_key": "zyx"}] is a better format just to make sure to avoid issues.

ilblackdragon:
@evgenykuzyakov I don't think anyone ever updated multisig contracts, honestly. 
We actually don't have a good way to remove multisig that would clean up it's storage either...

Are you suggesting that non upgradable between each other versions of contracts are spawned into separate cargo / repo?

multisig-factory/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+codegen-units = 1
+# Tell `rustc` to optimize for small code size.
+opt-level = "z"
+lto = true
+debug = false
+panic = "abort"
+overflow-checks = true
+
+[dependencies]
+near-sdk = "2.0.0"
```
mikedotexe, line 21: Yikes, this is old, but not as important for the factory to up to date I suppose

multisig-factory/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "multisig-factory"
+version = "0.1.0"
+authors = ["Near Inc <hello@near.org>"]
+edition = "2018"
+publish = false
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
mikedotexe, line 9: We can remove `rlib` for an easy reduction to code size:
https://www.near-sdk.io/reducing-contract-size/examples#removing-rlib-from-the-manifest

mikedotexe, line 9:
```suggestion
crate-type = ["cdylib"]
```

multisig-factory/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
+use near_sdk::json_types::Base58PublicKey;
+use near_sdk::serde::{Deserialize, Serialize};
+use near_sdk::serde_json::json;
+use near_sdk::{env, near_bindgen, AccountId, Promise};
+
+#[global_allocator]
+static ALLOC: near_sdk::wee_alloc::WeeAlloc<'_> = near_sdk::wee_alloc::WeeAlloc::INIT;
+
+const CODE: &[u8] = include_bytes!("../../multisig2/res/multisig.wasm");
```
mikedotexe, line 10:
```suggestion
const CODE: &[u8] = include_bytes!("../../multisig2/res/multisig2.wasm");
```
See:
https://github.com/near/core-contracts/pull/125/files#diff-563ebbdd4d50f30899977aebf36f53bcda8646c5bb5f5c6804c615b73a634f53R5
and manifest:
https://github.com/near/core-contracts/pull/125/files#diff-d40e917a59bc0f7f77b1e0e76ad1de4f7052fc19428233acdb578caedd69296cR2

multisig2/README.md
```diff
@@ -0,0 +1,271 @@
...
+```bash
+near view multisig.illia get_confirmations '{"request_id": 0}'
+```
+
+Total confirmations required for any request:
+```
+near view multisig.illia get_num_confirmations
+```
+
+### Upgrade given multisig with new code
```
mikedotexe, line 229: Darn good README 👏🏼

multisig2/src/lib.rs
```diff
@@ -0,0 +1,878 @@
...
+    /// Call function on behalf of this contract.
+    FunctionCall {
+        method_name: String,
+        args: Base64VecU8,
+        deposit: U128,
+        gas: U64,
+    },
+    /// Sets number of confirmations required to authorize requests.
+    /// Can not be bundled with any other actions or transactions.
+    SetNumConfirmations { num_confirmations: u32 },
```
mikedotexe, line 67: `u32` might be excessively high, but probably not a big deal

multisig2/src/lib.rs
```diff
@@ -0,0 +1,878 @@
...
+    requests: UnorderedMap<RequestId, MultiSigRequestWithSigner>,
+    /// All confirmations for active requests.
+    confirmations: LookupMap<RequestId, HashSet<String>>,
+    /// Number of requests per member.
+    num_requests_pk: LookupMap<String, u32>,
+    /// Limit number of active requests per member.
+    active_requests_limit: u32,
+}
+
+#[inline]
```
mikedotexe, line 135: cool 👌🏼

multisig2/src/lib.rs
```diff
@@ -0,0 +1,878 @@
...
+
+    /// Add request for multisig and confirm with the pk that added.
+    pub fn add_request_and_confirm(&mut self, request: MultiSigRequest) -> RequestId {
+        let request_id = self.add_request(request);
+        self.confirm(request_id);
+        request_id
+    }
+
+    /// Remove given request and associated confirmations.
+    pub fn delete_request(&mut self, request_id: RequestId) {
```
mikedotexe, line 210: Might not be a bad idea to return `MultiSigRequest` since `remove_request` will give us that anyway.
Might be useful for a frontend to say, "Deleted <some info>…"
Not strongly held, just a thought.

mikedotexe (approved):
Looks great except for the tiny modifications, mainly making sure the factory points to multisig2

multisig-factory/Cargo.toml
ilblackdragon: factory side doesn't matter much. and removing this then stops from working the integration tests.

PR #128, opened 2021-02-09
Add w-near nep-146, nep-145, nep-148

evgenykuzyakov:


ilblackdragon:
Moved the interfaces and basis of the implementation into https://github.com/near/near-sdk-rs/pull/293

wNEAR implementation can be just adding methods for wnear_deposit and wnear_withdraw
Plus IMO it should auto register accounts on wnear_deposit -- there is now method `internal_register_account` specifically for that.

evgenykuzyakov:
Closing in favor of #134

PR #129, opened 2021-02-10
Simple NFT

evgenykuzyakov:
Remaining items as I see it:
- [x] Use Storage Management standard because this uses enumeration, which adds to a new map.
- [ ] The `NonFungibleTokenCore` trait has mixed other standards in there, including approvals and enumeration. Let's separate these into different traits and corresponding files.


nft-simple/src/nft_core.rs
```diff
@@ -60,54 +78,121 @@ impl NonFungibleTokenCore for Contract {
...
             token_id,
             &env::current_account_id(),
             NO_DEPOSIT,
             GAS_FOR_RESOLVE_TRANSFER,
         ))
     }
 
+    #[payable]
+    fn nft_approve_account_id(&mut self, token_id: TokenId, account_id: ValidAccountId) -> bool {
+        let mut token = self.tokens_by_id.get(&token_id).expect("Token not found");
```
mikedotexe, line 137: I think this would be a likely attack vector and we'd want:

    assert_one_yocto();

nft-simple/src/nft_core.rs
```diff
@@ -20,35 +21,52 @@ pub trait NonFungibleTokenCore {
...
+
     fn nft_total_supply(&self) -> U64;
+
+    fn nft_token(&self, token_id: TokenId) -> Option<Token>;
 }
 
 #[ext_contract(ext_non_fungible_token_receiver)]
 trait NonFungibleTokenReceiver {
     /// Returns `true` if the token should be returned back to the sender.
     /// TODO: Maybe make it inverse. E.g. true to keep it.
```
mikedotexe, line 43: I was thinking about this one, too. I was wondering if we even need a return at all.
This is an interesting approach we can also consider from ERC-721:

>it calls `onERC721Received` on `_to` and throws if the return value is not `bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))`.

https://eips.ethereum.org/EIPS/eip-721

nft-simple/src/internal.rs
```diff
@@ -24,6 +27,37 @@ pub(crate) fn assert_self() {
...
+    );
+
+    let refund = attached_deposit - required_cost;
+    if refund > 0 {
+        Promise::new(env::predecessor_account_id()).transfer(refund);
+    }
+}
+
+pub(crate) fn bytes_for_approved_account_id(account_id: &AccountId) -> u64 {
+    account_id.len() as u64 + 4
```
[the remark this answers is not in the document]
evgenykuzyakov: Added comment

nft-simple/src/nft_core.rs
evgenykuzyakov: Basically, we should `panic` if we don't want to receive. I can see a scenario where this is undesirable. Let's say the receiver contract needs to do an extra cross-contract operation before it can finalize the transfer. In this case, the extra operation requires a callback back on the receiver contract. But if you have to panic at the callback, then you can't revert the state of the receiver. This probably can be avoided, but a the same time, explicit `true` is more secure.

nft-simple/src/nft_core.rs
evgenykuzyakov: Good catch

nft-simple/src/internal.rs
```diff
@@ -0,0 +1,155 @@
+use crate::*;
+
+/// Price per 1 byte of storage from mainnet config after `0.18` release and protocol version `42`.
```
bowenwang1996, line 3:
```suggestion
/// Price per 1 byte of storage from mainnet config after `1.18.0` release and protocol version `42`.
```

nft-simple/src/mint.rs
```diff
@@ -0,0 +1,27 @@
+use crate::*;
+
+#[near_bindgen]
+impl Contract {
+    #[payable]
+    pub fn nft_mint(&mut self, token_id: TokenId, metadata: String) {
+        let initial_storage_usage = env::storage_usage();
+        self.assert_owner();
+        let token = Token {
+            owner_id: self.owner_id.clone(),
```
mikedotexe, line 10: Feels like the `owner_id` would. be something we'd pass in as a parameter to `nft_mint(…)` to me

evgenykuzyakov: why?

mikedotexe:
This PR should be against `master` branch, yes?

nft-simple/src/internal.rs
```diff
@@ -1,5 +1,7 @@
 use crate::*;
 
+const YOCTO_MULTIPLIER: f32 = 0.000000000000000000000001;
```
bowenwang1996, line 3: why do we use floats?

mikedotexe: I wanted a clear error that's helpful to folks, is there a better way I should do this? Pitfalls or advice welcome.

Screen Shot 2021-03-26 at 9 16 04 AM

mikedotexe: removed

nft-simple/src/nft_core.rs
mikedotexe: Looks like this has been addressed

mikedotexe:
> Shall we rename reference => reference_url?

We're trying to stick with the merged spec on this one, thanks tho! It is slightly odd
https://nomicon.io/Standards/NonFungibleToken/Metadata.html

mikedotexe:
Updated the initial description of this PR with changes I think we need before merging.

mikedotexe:
Closing in favor of putting this in the contract standards in the sdk https://github.com/near/near-sdk-rs/pull/361

PR #131, opened 2021-02-13
added assertion that asserts that the amount being transfered is larger than 0

jasperdg:
There is no use-case for transferring 0 tokens to an account and it could cause overflows on receiving contract's end.


bowenwang1996 (approved):

evgenykuzyakov (changes requested):
Please remove other changes such as attaching storage

w-near-141/src/fungible_token_core.rs
```diff
@@ -171,7 +176,10 @@ impl FungibleTokenResolver for Contract {
                     amount
                 }
             }
-            PromiseResult::Failed => amount,
+            PromiseResult::Failed => {
+                Promise::new(sender_id.to_string()).transfer(initial_storage_deposit.into());
```
evgenykuzyakov, line 180: This can be abused, since the receiver can return a promise that fails, which doesn't generate a refund.

So this fungible contract core implementation can't be sure that there was refund.

It's a limitation of NEAR protocol cross-contracts right now.

jasperdg: Yup agreed, reset to `8da15fa`

evgenykuzyakov (commented):
Added change in https://github.com/near/core-contracts/commit/1240e83eaae4d97856842bf5f223e04eba8ef394

oysterpack:
The ['StorageManager][2] trait is not fully compliant with the proposed [standard interface][1].
- `amount` arg on `storage_withdraw` function should be optional:

```rust
fn storage_withdraw(&mut self, amount: Option<U128>) -> AccountStorageBalance; 
```

[1]: https://github.com/near/NEPs/discussions/145#discussioncomment-361351
[2]: https://github.com/near/core-contracts/blob/w-near-141/w-near-141/src/storage_manager.rs#L19

PR #134, opened 2021-02-25
Reimplement wNEAR with near-contract-standards

evgenykuzyakov:
NOTE: Legacy means the contract deployed on `wrap.near` (not NEP-21).

- Updates wNEAR  implementation to use FT standard. Exposes legacy storage method `storage_minimum_balance` which is available on `wrap.near` contract.
- Upgrade `rust-toolchain` to stable Rust `1.51.0`. Rebuild with contract builder.

## Test plan
- Add sim tests to test legacy->current migration
- Verified on testnet


rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2020-05-15
+stable-2020-10-08
```
bowenwang1996, line 1: if it is stable, why do we even need this file

bowenwang1996 (commented):
Did you forget to include some file?

rust-toolchain
evgenykuzyakov: It's the one we use for `near-sdk-rs`. Maybe people had old toolchains by default in the past.

bowenwang1996 (approved):

ilblackdragon:
Let's upgrade this to using near-standard-contracts and push?
We also need to hot swap the current version and update Flux

ilblackdragon:
My suggestion we keep the old storage API for Flux UI to work.

w-near-141/Cargo.toml
```diff
@@ -8,7 +8,10 @@ edition = "2018"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-near-sdk = "2.0.0"
+near-sdk = { git = "https://github.com/near/near-sdk-rs" }
```
willemneal, line 11: Should pin a commit until published.

w-near/src/lib.rs
```diff
@@ -1,943 +1,58 @@
...
-            attached_deposit: 0,
-            prepaid_gas: 10u64.pow(18),
-            random_seed: vec![0, 1, 2],
-            is_view: false,
-            output_data_receivers: vec![],
-            epoch_height: 0,
+near_contract_standards::impl_fungible_token_core!(Contract, ft);
+near_contract_standards::impl_fungible_token_storage!(Contract, ft);
+
+impl FungibleTokenMetadataProvider for Contract {
```
ilblackdragon, line 46: #[near_bindgen]

ilblackdragon (approved):
LGTM with one line missing to expose ft_metadata

PR #144, opened 2021-05-04
docs(lockup-contract): refactored README for better clarity

### LINKED ISSUE #140: Possible error in curent version of lockup contract
I'm not sure if this is intended, but it seems like there is an error in calculating the lockup schedule for the accounts with `lockup_duration`.

If you check here:
https://github.com/near/core-contracts/blob/master/lockup/src/getters.rs#L74
the `lockup_duration` is added to `lockup_timestamp`. Next the `lockup_timestamp ` is used to calculate `end_timestamp ` and then it's used to calculate `time_left`. `time_left` is used to calculate the `unreleased_amount`. 

If you consider example account:
 - "lockup_duration": "31536000000000000" (1 year)
 - "release_duration": "126144000000000000" (4 years)
 - "lockup_timestamp": null

The current version of `get_locked_amount` (lockup getters) will calculate:
 - `lockup_timestamp`  - 10/13/2021
 - `end_timestamp` - 10/12/2025
so the lockup schedule is moved one year ahead.

It will also calculate wrong values if the lockup has `lockup_duration` and has `lockup_timestamp`, as it will not include `lockup_duration` into calculations. But I don't know if these types of accounts will be ever created.

If we don't plan to add a new contract:
 -  without `lockup_timestamp` and with `lockup_duration`
 - or with `lockup_timestamp` and with `lockup_duration`
than this issue might be closed.

Just noticed this problem, but I'm not sure if it's relevant to future lockups.

### LINKED ISSUE #141: Cliff does not work as expected
Looks like we have a bug in calculation of [unrelease_amount](https://github.com/near/core-contracts/blob/master/lockup/src/getters.rs#L87):

```
let unreleased_amount = U256::from(lockup_amount) * time_left / U256::from(release_duration);
```

Let's suppose that we have classic contract 4-years long in total, with 1-year cliff.
It does not clear, what `release_duration` means. [Doc](https://github.com/near/core-contracts/blob/master/lockup/src/lib.rs#L173) says it's `the duration when the full lockup amount will be available`. Does it mean 3 or 4 years for our case? Anyway, we will have problems further.

1. Let's suppose that `release_duration` == 3 years for such case.
It means that `unreleased_amount` will not take into account cliff. Our `lockup_amount` will start unlocking after we pass the cliff without unlocking 25% at the early beginning. To fix that, the formula should look like:
```
let unreleased_amount = U256::from(lockup_amount) * time_left / U256::from(release_duration + self.lockup_information.lockup_duration);
```

2. Let's suppose that `release_duration` == 4 years for such case.
It means that our contract will last 5 years instead of 4 because of [this line](https://github.com/near/core-contracts/blob/master/lockup/src/getters.rs#L81). To fix that, we should **also** apply this change:
```
let end_timestamp = lockup_timestamp.saturating_add(release_duration - self.lockup_information.lockup_duration);
```
(code could be simplified, I just tried to explain the idea)

telezhnaya:
What have been done:
- Added info about the difference between lockup and vesting, and how do we combine these logic
- Added info about lockup fields
- Added info about foundation account (it could be trird-party)
- Removed `technical details` section. All the info from it is outdated (first point) or written in corresponding sections.

Resolves #140, resolves #141

### DIFF
--- a/lockup/README.md
+++ b/lockup/README.md
@@ -6,7 +6,7 @@ This contract acts as an escrow that locks and holds an owner's tokens for a loc
 The contract consists of lockup and vesting processes that go simultaneously.
 A high-level overview could be found [in NEAR documentation](https://docs.near.org/docs/tokens/lockup).
 
-A lockup period starts from the specified timestamp (or, if not specified, from the moment when [transfers were enabled by voting](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/) on 2020-10-13, 18:38:58 UTC or `1602614338293769340` nanoseconds unix time) and lasts for the specified duration.
+A lockup period starts from the specified timestamp and lasts for the specified duration.
 Tokens will be unlocked linearly.
 
 Vesting is an additional mechanism. It also locks the tokens, and it allows to configure 2 more options:
@@ -14,9 +14,9 @@ Vesting is an additional mechanism. It also locks the tokens, and it allows to c
 2. Cliff vesting period.
 
 The owner can add a full access key to the account if all conditions are met:
-- [Transfers are enabled](https://near.org/blog/near-mainnet-is-now-community-operated/) ([Phase II launch on MainNet enabled transfers](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/));
 - Vesting and lockup processes finished;
-- The termination process is ended (if applicable).
+- The termination process is ended (if applicable);
+- [Transfers are enabled](https://near.org/blog/near-mainnet-is-now-community-operated/) ([Phase II launch on MainNet enabled transfers](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/)).
 
 This will allow the owner to turn this contract account into a regular account, claim the remaining tokens, and remove the contract or delete the account.
 
@@ -25,18 +25,19 @@ This will allow the owner to turn this contract account into a regular account, 
 Lockup is a mechanism of linear unlocking of tokens that could not be terminated.
 It is described by the following fields:
 - `lockup_timestamp` - The moment when tokens start linearly unlocking;
-- `lockup_duration` - Alternative way to set the moment when the tokens become unlock.
-  The duration from the moment transfers were enabled to the moment when linear unlocking begins;
+- `lockup_duration` - [deprecated] Alternative way to set the moment when the tokens become unlock.
+  The duration from [the moment transfers were enabled](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/) to the moment when linear unlocking begins;
 - `release_duration` - The length of the unlocking schedule during which tokens are linearly unlocked.
   By the end of this duration all tokens are unlocked.
   `finish_timestamp = lockup_timestamp + release_duration`.
 
+If `lockup_timestamp` and `lockup_duration` are not specified, the lockup starts from the timestamp from [`transfers_information`](https://github.com/near/core-contracts/blob/master/lockup/src/lib.rs#L187) field.
+It's usually the moment when [transfers were enabled by voting](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/) in the system: 2020-10-13, 18:38:58 UTC or `1602614338293769340` nanoseconds unix time.
+
 ### Vesting schedule
 
 The contract can contain a vesting schedule and serve as a vesting agreement between the foundation and an employee (owner of the contract).
-
-Foundation is either the NEAR Foundation, or it could be a third-party account.
-It is set at the moment of initializing the contract by the `foundation_account_id` field.
+The foundation is set at the moment of initializing the contract by the `foundation_account_id` field.
 
 A vesting schedule is described by three timestamps in nanoseconds:
 - `start_timestamp` - When the vesting starts. E.g. the start date of employment;
@@ -44,7 +45,10 @@ A vesting schedule is described by three timestamps in nanoseconds:
   The remaining tokens will vest continuously until they are fully vested.
   Assume we have a 4-year contract with a 1-year cliff.
   In the first year, nothing is vested, then 25% is vested, then we have linear vesting till the end of the contract.
-  25% is calculated by the formula `(cliff_timestamp - start_timestamp) / (end_timestamp - start_timestamp)`;
+  25% is the number calculated by the formula:
+  ```
+  cliff_tokens_percentage = (cliff_timestamp - start_timestamp) / (end_timestamp - start_timestamp)
+  ```
 - `end_timestamp` -  When the vesting ends.
 
 Once the `cliff_timestamp` passed, the tokens are vested on a pro-rata basis from the `start_timestamp` to the `end_timestamp`.
@@ -53,15 +57,18 @@ Once the `cliff_timestamp` passed, the tokens are vested on a pro-rata basis fro
 
 The contract could have both lockup and vesting schedules.
 
-The tokens are started to become liquid at the following timestamp `max(max(transfers_enabled_timestamp + lockup_duration, lockup_timestamp), cliff_timestamp)`.
+The tokens start to become liquid at the timestamp:
+```
+liquidity_timestamp = max(max(transfers_enabled_timestamp + lockup_duration, lockup_timestamp), cliff_timestamp)
+```
 
 The current amount of non-liquid tokens are calculated as the maximum between lockup and vesting logic.
 If at least one mechanism said the tokens are locked, then they are still locked.
 
 The contract could also have only one of these mechanisms.
 When initializing, it's possible to pass empty vesting information, then we use a lockup schedule.
-It's also possible not to provide `release_duration`, it means that we are using a vesting schedule.
-If neither of the mechanisms is initialized, the tokens will be locked till the transfers enabled moment that set up in the account.
+It's also possible not to provide `release_duration`, it means that we use a vesting schedule.
+If neither of the mechanisms is initialized, the tokens will become liquid after transfers enabled moment ([`transfers_information`](https://github.com/near/core-contracts/blob/master/lockup/src/lib.rs#L187) field).
 
 ### Staking
 
@@ -71,7 +78,6 @@ This contract doesn't allow to directly stake from this account, so the owner ca
 
 The owner can choose the staking pool for delegating tokens.
 The staking pool contract and the account have to be approved and whitelisted by the NEAR Foundation, to prevent lockup tokens from being lost, locked, or stolen.
-This staking pool must be an approved account, which is validated by a whitelisting contract.
 Once the staking pool holds tokens, the owner of the staking pool can use them to vote on the network governance issues, such as enabling transfers.
 So the owner needs to pick the staking pool that fits the best.
 
@@ -88,22 +94,6 @@ Once the unvested balance is withdrawn completely, the contract returns to the r
 
 The amount withdrawn in the event of termination by the foundation may be lower than the initial contract amount.
 It's because the contract has to maintain the minimum required balance to cover storage of the contract code and contract state.
-The amount of NEAR tokens locked to maintain the minimum storage balance is `35` NEAR.
-`35` NEAR is enough to cover storage for `350000` bytes on Mainnet at the price of `1` NEAR per `10000` bytes.
-
-If there is still a termination balance deficit due to the minimum required balance, the owner may decide to fund the deficit on this account to finish the termination process.
-This can be done through a regular transfer action from an account with a liquid balance.
-
-### Private vesting schedule
-
-Since the vesting schedule usually starts at the date of employment it allows to de-anonymize the owner of the lockup contract.
-To keep the identity private, the contract allows to hash the vesting schedule with some random salt and keep store the hash instead of the raw vesting schedule information.
-In case the foundation has to terminate the vesting schedule, it will provide the raw vesting schedule and the salt, effectively revealing the vesting schedule.
-The contract then will compare the hash with the internal hash and if they match proceed with the termination.
-
-NOTE: The private vesting schedule can only be used if the lockup release period and the lockup duration are effectively shadowing the vesting duration.
-Meaning that the lockup release ends later than the vesting release and the lockup duration ends after the vesting cliff.
-Once the lockup schedule starts before the vesting schedule (e.g. employment starts after the transfers are enabled), the vesting schedule can't be kept private anymore.
 
 ### Guarantees
 
@@ -113,6 +103,17 @@ With the guarantees from the staking pool contracts, whitelist, and voting contr
 - The owner can not withdraw tokens locked due to lockup period, disabled transfers, or vesting schedule.
 - The owner can withdraw rewards from the staking pool before tokens are unlocked unless the vesting termination prevents it.
 - The owner should be able to add a full access key to the account, once all tokens are vested, unlocked and transfers are enabled.
+
+### [Deprecated] Private vesting schedule
+
+Since the vesting schedule usually starts at the date of employment it allows to de-anonymize the owner of the lockup contract.
+To keep the identity private, the contract allows to hash the vesting schedule with some random salt and keep store the hash instead of the raw vesting schedule information.
+In case the foundation has to terminate the vesting schedule, it will provide the raw vesting schedule and the salt, effectively revealing the vesting schedule.
+The contract then will compare the hash with the internal hash and if they match proceed with the termination.
+
+**NOTE**: The private vesting schedule can only be used if the lockup release period and the lockup duration are effectively shadowing the vesting duration.
+Meaning that the lockup release ends later than the vesting release and the lockup duration ends after the vesting cliff.
+Once the lockup schedule starts before the vesting schedule (e.g. employment starts after the transfers are enabled), the vesting schedule can't be kept private anymore.
 
 ## Change Log
 
--- a/lockup/README.md
+++ b/lockup/README.md
@@ -1,107 +1,117 @@
 # Lockup / Vesting contract
 
-# PLEASE WAIT UNTIL [ISSUE 141](https://github.com/near/core-contracts/issues/141) IS RESOLVED BEFORE DEPLOYING THE CONTRACT
-
 ## Overview
 
-This contract acts as an escrow that locks and holds an owner's tokens for a lockup period. A lockup period starts
-from the moment transfers are enabled by voting and lasts for the specified duration.
-It's also possible to lock tokens until the absolute timestamp. In this case the tokens will be available after the
-lockup period or after the timestamp, whichever is later.
+This contract acts as an escrow that locks and holds an owner's tokens for a lockup period.
+The contract consists of lockup and vesting processes that go simultaneously.
+A high-level overview could be found [in NEAR documentation](https://docs.near.org/docs/tokens/lockup).
 
-If transfers are not enabled yet, the contract keeps the account ID of the transfer poll contract.
-When the transfer poll is resolved, it returns the timestamp when it was resolved and it's used as the beginning of the
-lockup period.
+A lockup period starts from the specified timestamp (or, if not specified, from the moment when [transfers were enabled by voting](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/) on 2020-10-13, 18:38:58 UTC or `1602614338293769340` nanoseconds unix time) and lasts for the specified duration.
+Tokens will be unlocked linearly.
 
-Once all tokens are unlocked, everything is vested, and transfers are enabled, the owner can add a full access key to the
-account. This will allow the owner to turn this contract account into a regular account, claim the remaining tokens, and remove the contract
-or delete the account.
+Vesting is an additional mechanism. It also locks the tokens, and it allows to configure 2 more options:
+1. Ability to terminate tokens vesting and refund non-vested tokens back.
+2. Cliff vesting period.
+
+The owner can add a full access key to the account if all conditions are met:
+- [Transfers are enabled](https://near.org/blog/near-mainnet-is-now-community-operated/) ([Phase II launch on MainNet enabled transfers](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/));
+- Vesting and lockup processes finished;
+- The termination process is ended (if applicable).
+
+This will allow the owner to turn this contract account into a regular account, claim the remaining tokens, and remove the contract or delete the account.
+
+### Lockup schedule
+
+Lockup is a mechanism of linear unlocking of tokens that could not be terminated.
+It is described by the following fields:
+- `lockup_timestamp` - The moment when tokens start linearly unlocking;
+- `lockup_duration` - Alternative way to set the moment when the tokens become unlock.
+  The duration from the moment transfers were enabled to the moment when linear unlocking begins;
+- `release_duration` - The length of the unlocking schedule during which tokens are linearly unlocked.
+  By the end of this duration all tokens are unlocked.
+  `finish_timestamp = lockup_timestamp + release_duration`.
 
 ### Vesting schedule
 
-The contract can contain a vesting schedule and serve as a vesting agreement between the NEAR Foundation (Foundation) and an employee (owner of contract).
-Vesting schedule is described by three timestamps in nanoseconds:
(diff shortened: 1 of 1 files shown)


lockup/README.md
```diff
@@ -1,43 +1,64 @@
 # Lockup / Vesting contract
 
-# PLEASE WAIT UNTIL [ISSUE 141](https://github.com/near/core-contracts/issues/141) IS RESOLVED BEFORE DEPLOYING THE CONTRACT
-
 ## Overview
 
-This contract acts as an escrow that locks and holds an owner's tokens for a lockup period. A lockup period starts
-from the moment transfers are enabled by voting and lasts for the specified duration.
+This contract acts as an escrow that locks and holds an owner's tokens for a lockup period. 
+The contract contains of lockup and vesting processes that go simultaneously.
```
bowenwang1996, line 6: "contains of" is not correct English grammar. You might have meant "consists of"

lockup/README.md
```diff
@@ -1,43 +1,64 @@
...
+
+A lockup period starts from the specified timestamp (or, if not specified, from moment when transfers are enabled by voting) 
+and lasts for the specified duration.
```
bowenwang1996, line 9: As suggested by @matklad, we start to use one sentence per line https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

lockup/README.md
```diff
@@ -1,43 +1,64 @@
...
+- `lockup_duration` - Alternative way to set the moment when the tokens become to unlock. 
+  The duration from the moment the transfers are enabled (`1602614338293769340`) to start of the contract.
+- `release_duration` - Length of the contract. `finish_timestamp = lockup_timestamp + release_duration`
+
 ### Vesting schedule
 
-The contract can contain a vesting schedule and serve as a vesting agreement between the NEAR Foundation (Foundation) and an employee (owner of contract).
+The contract can contain a vesting schedule and serve as a vesting agreement between the Foundation and an employee (owner of contract).
+
+Foundation is either the NEAR Foundation, or it could be third-party account.
```
bowenwang1996, line 35: if so, is `foundation` still the best name?

lockup/README.md
```diff
@@ -1,43 +1,64 @@
...
 ## Overview
 
-This contract acts as an escrow that locks and holds an owner's tokens for a lockup period. A lockup period starts
-from the moment transfers are enabled by voting and lasts for the specified duration.
+This contract acts as an escrow that locks and holds an owner's tokens for a lockup period. 
+The contract contains of lockup and vesting processes that go simultaneously.
+
+A lockup period starts from the specified timestamp (or, if not specified, from moment when transfers are enabled by voting) 
+and lasts for the specified duration.
 It's also possible to lock tokens until the absolute timestamp. In this case the tokens will be available after the
```
frol, line 10: > It's also possible to lock tokens until the absolute timestamp.

I don't believe that is true.

lockup/README.md
```diff
@@ -1,43 +1,64 @@
...
 lockup period or after the timestamp, whichever is later.
 
-If transfers are not enabled yet, the contract keeps the account ID of the transfer poll contract.
-When the transfer poll is resolved, it returns the timestamp when it was resolved and it's used as the beginning of the
-lockup period.
+Vesting is an additional mechanism. It also locks the tokens, and it allows to configure 2 more options:
+1. Ability of terminating the contract from the Foundation.
```
frol, line 14:
```suggestion
1. Ability to terminate tokens vesting and refund non-vested tokens back.
```

lockup/README.md
```diff
@@ -1,43 +1,64 @@
...
+2. Cliff: the period when all tokens remain locked. After the cliff, there is a linear unlock till the end of the contract.
```
frol, line 15: Maybe something like this:

```suggestion
2. Cliff vesting period. All tokens remain non-vested before the cliff checkpoint is reached; on the cliff checkpoint, the accumulated tokens are vested in bulk; after the cliff, tokens are linearly unlocked till the vesting reaches its end timestamp.
```

or, just:

```suggestion
2. Cliff vesting period.
```

(the details are provided below)

lockup/README.md
```diff
@@ -1,43 +1,64 @@
...
+The tokens are started to unlock at the following timestamp `max(lockup_timestamp, cliff_timestamp)`.
 NOTE: If the `lockup_timestamp` is not specified, the tokens are unlocked at `transfers_enabled_timestamp + lockup_duration`.
 
-Release schedule can not be terminated by the foundation.
+Current amount of locked tokens are calculated as the maximum between lockup and vesting logic.
+If at least one mechanism said the tokens are locked, then they are still locked.
+
+The contract could also have only one of these mechanisms. When initializing, it's possible to pass empty vesting information. 
+It's also possible not to provide `release_duration`, it means that we are using vesting schedule.
+If both mechanisms are not initialized, the tokens will be locked forever.
```
frol, line 61:
```suggestion
If neither of the mechanisms is initialized, the tokens will be locked forever.
```

telezhnaya:
I also created additional doc with cute pictures: https://github.com/telezhnaya/docs/blob/master/docs/tokens/lockup_vesting_details.md

lockup/README.md
frol: @telezhnaya `foundation_account_id` is an unfortunate internal naming, but we still can use better wording on the documentation level and reference the internal naming explicitly pointing out where it comes from.

frol (commented):
Overall, looks accurate to me. Once we have the visualizations, it will be perfect

frol:
BTW, I suggest using https://grammarly.com/ to check the grammar :wink:

lockup/README.md
telezhnaya: I think it's true, you could set `lockup_timestamp` to desired absolute timestamp and set `release_duration` to 0.
But it looks like this sentence adds nothing to the meaning, it could be deleted not to confuse anyone

lockup/README.md
matklad: @bowenwang1996 do we have some style.md somewhere where we could dump this info? A-la https://github.com/rust-analyzer/rust-analyzer/blob/d4a51241ceedae25f15bb8d202493554ae0af810/docs/dev/style.md#documentation ?

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
-In addition to the lockup period that starts from the moment the transfers are enabled, vesting schedule also locks
-all tokens until `cliff_timestamp`.
-Once the `cliff_timestamp` passed, the tokens are vested on a pro rata basis from the `start_timestamp` to the `end_timestamp`.
+Once the `cliff_timestamp` passed, the tokens are vested on a pro-rata basis from the `start_timestamp` to the `end_timestamp`.
 
-### Release schedule
+### Combining lockup and vesting
 
-The release of tokens can also be scheduled to be linear from the moment tokens are unlocked.
-The tokens are unlocked at the following timestamp `max(transfers_enabled_timestamp + lockup_duration, lockup_timestamp)`.
```
evgenykuzyakov, line 37: Why this is removed? If both `lockup_duration` and `lockup_timestamp` is present, then we still should use the maximum of two. So with `cliff_timestamp`, we have:
```
max(max(transfers_enabled_timestamp + lockup_duration, lockup_timestamp), cliff_timestamp)
```

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
-Once all tokens are unlocked, everything is vested, and transfers are enabled, the owner can add a full access key to the
-account. This will allow the owner to turn this contract account into a regular account, claim the remaining tokens, and remove the contract
-or delete the account.
+### Lockup schedule
+
+Lockup is a mechanism of linear unlocking of tokens that could not be terminated.
+It is described by the following fields:
+- `lockup_timestamp` - The moment when the tokens become unlock (start of the contract).
+- `lockup_duration` - Alternative way to set the moment when the tokens become unlock.
+  The duration from the moment the transfers are enabled (`1602614338293769340`) to the start of the contract.
```
evgenykuzyakov, line 24: `TransferEnabled` timestamp can also be passed directly, so it might be different.

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
 
-# PLEASE WAIT UNTIL [ISSUE 141](https://github.com/near/core-contracts/issues/141) IS RESOLVED BEFORE DEPLOYING THE CONTRACT
-
 ## Overview
 
-This contract acts as an escrow that locks and holds an owner's tokens for a lockup period. A lockup period starts
-from the moment transfers are enabled by voting and lasts for the specified duration.
-It's also possible to lock tokens until the absolute timestamp. In this case the tokens will be available after the
-lockup period or after the timestamp, whichever is later.
+This contract acts as an escrow that locks and holds an owner's tokens for a lockup period.
```
frol, line 5: Let's add a link to https://docs.near.org/docs/tokens/lockup somewhere

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
-The amount withdrawn in the event of termination by NEAR foundation may be lower than the initial contract amount.
+The amount withdrawn in the event of termination by Contract Creator may be lower than the initial contract amount.
 It's because the contract has to maintain the minimum required balance to cover storage of the contract code and contract state.
 The amount of NEAR tokens locked to maintain the minimum storage balance is `35` NEAR.
-`35` NEAR is enough to cover storage for `350000` bytes on mainnet at price of `1` NEAR per `10000` bytes.
+`35` NEAR is enough to cover storage for `350000` bytes on Mainnet at the price of `1` NEAR per `10000` bytes.
 
-If there is still a termination balance deficit due to minimum required balance, the owner may decide to fund the deficit on this account to finish the termination process.
-This can be done through a regular transfer action from an account with liquid balance.
+If there is still a termination balance deficit due to the minimum required balance, the owner may decide to fund the deficit on this account to finish the termination process.
```
nearmax, line 87: This sounds like an important mechanism that everyone should be aware of. Is it possible to pre-fund all lockup contracts upon creation with 35 extra NEAR to avoid having this situation upon termination?

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
-
-The contract can be used for the following purposes:
-- Lock tokens until the transfers are voted to be enabled.
-- Lock tokens for the lockup period and until the absolute timestamp, whichever is later.
-- Lock tokens for the lockup period without a vesting schedule. All tokens will be unlocked at once once the lockup period passed.
-- Lock tokens for the lockup period with a vesting schedule.
-  - If the NEAR Foundation account ID is provided during initialization, the NEAR Foundation can terminate vesting schedule.
-  - If the NEAR Foundation account ID is not provided, the vesting schedule can't be terminated.
-- Lock tokens for the lockup period with the release duration. Tokens are linearly released on transfers are enabled.
+Since the vesting schedule usually starts at the date of employment it allows to de-anonymize the owner of the lockup contract.
```
nearmax, line 92: To clarify, this contract will be used not just for the employees.

nearmax (changes requested):

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
+This will allow the owner to turn this contract account into a regular account, claim the remaining tokens, and remove the contract or delete the account.
 
-Once all tokens are unlocked, everything is vested, and transfers are enabled, the owner can add a full access key to the
-account. This will allow the owner to turn this contract account into a regular account, claim the remaining tokens, and remove the contract
-or delete the account.
+### Lockup schedule
+
+Lockup is a mechanism of linear unlocking of tokens that could not be terminated.
+It is described by the following fields:
+- `lockup_timestamp` - The moment when the tokens become unlock (start of the contract).
```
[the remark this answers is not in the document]
telezhnaya: Sorry for the confusing phrasing.
Took your version except example in p.1. The idea of the timestamp (as I see it) is that we can forget about Mainnet launch, so I don't think this example is relevant right now.

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
- Example: a 1 year of employment at which moment the 1/4 of tokens become vested.
+  The remaining tokens will vest continuously until they are fully vested.
+  Assume we have a 4-year contract with a 1-year cliff.
+  In the first year, nothing is vested, then 25% is vested, then we have linear vesting till the end of the contract.
 - `end_timestamp` -  When the vesting ends.
 
-In addition to the lockup period that starts from the moment the transfers are enabled, vesting schedule also locks
-all tokens until `cliff_timestamp`.
-Once the `cliff_timestamp` passed, the tokens are vested on a pro rata basis from the `start_timestamp` to the `end_timestamp`.
+Once the `cliff_timestamp` passed, the tokens are vested on a pro-rata basis from the `start_timestamp` to the `end_timestamp`.
```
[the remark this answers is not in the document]
telezhnaya: Are you sure that example above is not enough? (5 lines before)
Don't want to overload the doc.

We also have many examples [here](https://github.com/telezhnaya/docs/blob/master/docs/tokens/lockup_vesting_details.md), it will be merged into [our doc](https://docs.near.org/docs/tokens/lockup), I think we don't need additional example here.

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
 
-### Release schedule
+### Combining lockup and vesting
 
-The release of tokens can also be scheduled to be linear from the moment tokens are unlocked.
-The tokens are unlocked at the following timestamp `max(transfers_enabled_timestamp + lockup_duration, lockup_timestamp)`.
+The contract could have both lockup and vesting schedules.
+
+The tokens are started to unlock at the following timestamp `max(max(transfers_enabled_timestamp + lockup_duration, lockup_timestamp), cliff_timestamp)`.
```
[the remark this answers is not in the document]
telezhnaya: They work separately, but the resulting schedule is affected by both these processes.

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
 NOTE: If the `lockup_timestamp` is not specified, the tokens are unlocked at `transfers_enabled_timestamp + lockup_duration`.
 
-Release schedule can not be terminated by the foundation.
+The current amount of locked tokens are calculated as the maximum between lockup and vesting logic.
+If at least one mechanism said the tokens are locked, then they are still locked.
+
+The contract could also have only one of these mechanisms. When initializing, it's possible to pass empty vesting information.
```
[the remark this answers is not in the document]
telezhnaya: `vesting_schedule` is optional field https://github.com/telezhnaya/core-contracts/blob/master/lockup/src/lib.rs#L188
just pass `None`

> Does it mean that we only have lockup mechanics?

Yes

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
+It's also possible not to provide `release_duration`, it means that we are using a vesting schedule.
+If neither of the mechanisms is initialized, the tokens will be locked forever.
```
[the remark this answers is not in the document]
telezhnaya: Great catch, thank you! Rewriting this

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
 
 ### Early Vesting Termination
 
-In case of vesting schedule, the contract supports the ability for the NEAR Foundation to terminate vesting at any point before it completes.
-If the vesting is terminated before the cliff all tokens are refunded to the Foundation. Otherwise the remaining unvested tokens are refunded.
+In the case of vesting schedule, the contract supports the ability for the Contract Creator to terminate vesting at any point before it completes.
+If the vesting is terminated before the cliff all tokens are refunded to the Contract Creator. Otherwise, the remaining unvested tokens are refunded.
 
-In the event of termination, the vesting stops and the remaining unvested tokens are locked until they are withdrawn by the Foundation.
+In the event of termination, the vesting stops, and the remaining unvested tokens are locked until they are withdrawn by the Contract Creator.
```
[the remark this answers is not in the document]
telezhnaya: I can change it but I try to be not too wordy. I think that my wording gives the same amount of information, doesn't it?

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
 During termination, the owner can't issue any action towards the staking pool or issue transfers.
-If the amount of tokens on the contract account is less than the remaining unvested balance, the Foundation will try to unstake and withdraw everything from the staking pool.
-Once the tokens are withdrawn from the staking pool, the Foundation will proceed with withdrawing the unvested balance from the contract.
+If the amount of tokens on the contract account is less than the remaining unvested balance, the Contract Creator will try to unstake and withdraw everything from the staking pool.
```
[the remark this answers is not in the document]
telezhnaya: It is possible if tokens are on the staking pool

lockup/README.md
```diff
@@ -1,107 +1,117 @@
...
+
+Vesting is an additional mechanism. It also locks the tokens, and it allows to configure 2 more options:
+1. Ability to terminate tokens vesting and refund non-vested tokens back.
+2. Cliff vesting period.
 
-If transfers are not enabled yet, the contract keeps the account ID of the transfer poll contract.
-When the transfer poll is resolved, it returns the timestamp when it was resolved and it's used as the beginning of the
-lockup period.
+The owner can add a full access key to the account if all conditions are met:
+- Transfers are enabled;
```
frol, line 17: This way it is much easier to comprehend :+1: 

I suggest providing more context whenever possible, so here we can provide the links to the initial voting plan and clarify that is already done:

```suggestion
- [Transfers are enabled](https://near.org/blog/near-mainnet-is-now-community-operated/) ([Phase II launch on MainNet enabled transfers](https://near.org/blog/near-mainnet-phase-2-unrestricted-decentralized/));
```

lockup/README.md
```diff
@@ -1,107 +1,117 @@
...
-
 ## Overview
 
-This contract acts as an escrow that locks and holds an owner's tokens for a lockup period. A lockup period starts
-from the moment transfers are enabled by voting and lasts for the specified duration.
-It's also possible to lock tokens until the absolute timestamp. In this case the tokens will be available after the
-lockup period or after the timestamp, whichever is later.
+This contract acts as an escrow that locks and holds an owner's tokens for a lockup period.
+The contract consists of lockup and vesting processes that go simultaneously.
+High-level overview could be found [in Near Docs](https://docs.near.org/docs/tokens/lockup).
```
frol, line 7:
```suggestion
A high-level overview could be found [in NEAR documentation](https://docs.near.org/docs/tokens/lockup).
```

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
-Once the staking pool holds tokens, the owner of the staking pool is able to use them to vote on the network governance issues, such as enabling transfers.
-So it's important for the owner to pick the staking pool that fits the best.
+Once the staking pool holds tokens, the owner of the staking pool can use them to vote on the network governance issues, such as enabling transfers.
+So the owner needs to pick the staking pool that fits the best.
 
 ### Early Vesting Termination
 
-In case of vesting schedule, the contract supports the ability for the NEAR Foundation to terminate vesting at any point before it completes.
-If the vesting is terminated before the cliff all tokens are refunded to the Foundation. Otherwise the remaining unvested tokens are refunded.
+In the case of vesting schedule, the contract supports the ability for the Contract Creator to terminate vesting at any point before it completes.
```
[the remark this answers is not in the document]
frol: I believe the confusion here is that "Contract Creator" is ambiguous since on mainnet it is lockup-factory which creates and deploys the contract. My suggestion would be to rename it to "foundation" (without "NEAR" and lowercase) implying any foundation which might be interested in founding lockup contracts and thus become the entity ensuring the trust to terminate vesting early. Our error messages in the contract return messages like:

> "Foundation account should be added for vesting schedule"

> "Foundation account can't be added without vesting schedule"

> "Can only be called by NEAR Foundation"

(the last one is unfortunately too specific)

lockup/README.md
telezhnaya: https://github.com/near/core-contracts/issues/145

telezhnaya:
Some changes were done. I cleaned up the history of commits, I also deleted changes from the code (I fixed some comments there, me and @frol decided not to touch the source code in this PR). 

The first commit is the original version + corrections after all reviews. In the second commit, I re-ordered things a little and marked `lockup_duration` and private vesting as deprecated.

frol (approved):
With this change and the change to the high-level overview in the NEAR docs (https://github.com/near/docs/pull/691), I believe, we reached the following:

* [x] Confirmed that the implementation itself is correct
* [x] The descriptions that were too low-level (reference-like) were extended with more context and references to the high-level ideas
* [x] The descriptions that were too generic (leading to uncertainty) were extended with specific examples and illustrations

@telezhnaya Thanks for the thorough review of all the information we had and driving this initiative forward!

bowenwang1996 (approved):

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
-The owner may want to stake these tokens (including locked tokens) to help secure the network and also earn staking rewards that are distributed to network validator.
+NEAR is the proof of stake network. The owner of the lockup contract might hold a large percentage of the network tokens.
+The owner may want to stake these tokens (including locked tokens) to help secure the network and also earn staking rewards that are distributed to the network validator.
 This contract doesn't allow to directly stake from this account, so the owner can delegate tokens to a staking pool contract (see https://github.com/near/initial-contracts/tree/master/staking-pool).
 
-The owner can choose the staking pool where to delegate tokens.
-The staking pool contract and the account has to be approved and whitelisted by the foundation, to prevent lockup tokens from being lost, locked or stolen.
+The owner can choose the staking pool for delegating tokens.
+The staking pool contract and the account have to be approved and whitelisted by the NEAR Foundation, to prevent lockup tokens from being lost, locked, or stolen.
```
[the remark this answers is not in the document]
nearmax: > It looks so. @evgenykuzyakov please confirm

Waiting for confirmation from @telezhnaya  or @evgenykuzyakov 

> I am not sure but it looks like it's not possible to pass the address of the whitelisting contract

It would be good if we could figure it out from the contract code. If whitelisting contract is hardcoded, then it would be good to have a link to the code where it is defined.

lockup/README.md
```diff
@@ -1,107 +1,110 @@
...
 This staking pool must be an approved account, which is validated by a whitelisting contract.
-Once the staking pool holds tokens, the owner of the staking pool is able to use them to vote on the network governance issues, such as enabling transfers.
-So it's important for the owner to pick the staking pool that fits the best.
+Once the staking pool holds tokens, the owner of the staking pool can use them to vote on the network governance issues, such as enabling transfers.
```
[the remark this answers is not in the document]
nearmax: Is it hardcoded in the contract? Could we please link the place from the source code in the documentation?

nearmax (approved):
Thanks for the documentation fix, approving assuming the remaining comments are resolved.

lockup/README.md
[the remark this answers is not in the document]
evgenykuzyakov: > The staking pool contract and the account have to be approved and whitelisted by the NEAR Foundation, to prevent lockup tokens from being lost, locked, or stolen.

This is guaranteed by the whitelist contract account ID. It doesn't have to be chosen by NEAR Foundation, but if reused for other lockups, the guarantee stands.

This question is completely independent from the `foundation_account_id` meaning.
The `foundation_account_id` can be renamed to `terminator_account_id`, because the only function of this account is to be able to terminate vesting schedule.

lockup/README.md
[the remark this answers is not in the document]
evgenykuzyakov: I think this part is obsolete and should be deprecated. Since the voting for transfers has already passed there is no reason for the new lockup contracts to be initialized without explicitly enabled transfers with the fixed timestamp.

The account_id that was checking and counting votes is not hardcoded, but a part of the enum:
https://github.com/near/core-contracts/blob/d84f7a924b79009d3f7aa4e577c07b08a0489dc9/lockup/src/types.rs#L62

I think we shouldn't use this anymore and should explicitly mark it as deprecated in the code.

frol:
I had to merge ASAP since it is good enough now, and I wanted to remove the scary banner and let partners review the contract without misleading messages. @telezhnaya please, submit a separate PR addressing the concerns from the comments

lockup/README.md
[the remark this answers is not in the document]
telezhnaya: I created a separate issue about marking as deprecated/fixing comment in the contract: https://github.com/near/core-contracts/issues/146
Please comment there anything that you remember that should be fixed

telezhnaya:
@frol thank you!
I continued [here](https://github.com/near/core-contracts/pull/147)

PR #148, opened 2021-06-15
Adding generic factory

ilblackdragon:
This is factory that allows to deploy any contracts that are stored inside it.

Anyone can store a contract and anyone else can use hash to deploy it with various parameters and calling initializer.


generic-factory/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+            access_keys,
+            method_name,
+            args.map(|args| args.0),
+        );
+    }
+}
+
+/// Stores attached data into blob store and returns hash of it.
+/// Implemented to avoid loading the data into WASM for optimal gas usage.
+#[cfg(target_arch = "wasm32")]
```
bowenwang1996, line 60: should we consider exposing some better API for this?

bowenwang1996 (commented):
Shouldn't we check attached_deposit can at least cover storage for the contract?

generic-factory/src/lib.rs
ilblackdragon: we should, but this is outside of this PR

generic-factory/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+            // Return from function value of register 1.
+            let blob_hash_str = near_sdk::serde_json::to_string(&Base58CryptoHash::from(blob_hash))
+                .unwrap()
+                .into_bytes();
+            b.borrow()
+                .as_ref()
+                .expect(BLOCKCHAIN_INTERFACE_NOT_SET_ERR)
+                .value_return(blob_hash_str.len() as _, blob_hash_str.as_ptr() as _);
+        });
+    }
```
austinabell, line 118:
```suggestion
    BLOCKCHAIN_INTERFACE.with(|b| unsafe {
        let interface = b.borrow().as_ref().expect(BLOCKCHAIN_INTERFACE_NOT_SET_ERR);
        // Load input into register 0.
        interface.input(0);
        // Compute sha256 hash of register 0 and store in 1.
        interface.sha256(u64::MAX as _, 0 as _, 1);
        // Check if such blob already stored.
        assert_eq!(
            interface.storage_has_key(u64::MAX as _, 1 as _),
            0,
            "ERR_ALREADY_EXISTS"
        );
        // Get length of the input argument and check that enough $NEAR has been attached.
        let blob_len = interface.register_len(0);
        let storage_cost = ((blob_len + 32) as u128) * env::storage_byte_cost();
        assert!(
            env::attached_deposit() >= storage_cost,
            "ERR_NOT_ENOUGH_DEPOSIT:{}",
            storage_cost
        );
        // Store value of register 0 into key = register 1.
        interface.storage_write(u64::MAX as _, 1 as _, u64::MAX as _, 0 as _, 2);
        // Load register 1 into blob_hash and save into LookupMap.
        let blob_hash = [0u8; 32];
        interface.read_register(1, blob_hash.as_ptr() as _);
        // Return from function value of register 1.
        let blob_hash_str = near_sdk::serde_json::to_string(&Base58CryptoHash::from(blob_hash))
            .unwrap()
            .into_bytes();
        interface.value_return(blob_hash_str.len() as _, blob_hash_str.as_ptr() as _);
    });
```

You can simplify these calls so it's a bit easier to read if you want, and it will make things easier to migrate to the new interface when available.

I can provide suggestions for the other usages if you'd like :)

generic-factory/src/lib.rs
austinabell: https://github.com/near/near-sdk-rs/pull/417 has been sitting for about a month, I can try to push for finalizing scope of the next release and getting there soon

austinabell: actually, you wouldn't need to wait for any release, can just define the interface that the `BLOCKCHAIN_INTERFACE` goes through and it should reduce contract size.

I'll do this now so there is an example, and any workflow to test the functionality would be appreciated

austinabell: https://github.com/near/core-contracts/pull/149 figured PR would be the easiest way to see changes. And the sys interface wasn't needed to be duplicated as it's included in `near_bindgen` right now.

PR #149, opened 2021-06-16
refactor(generic-factory): remove need to go through BLOCKCHAIN_INTERFACE for wasm env

austinabell:
Reduces contract size by a few thousand bytes on my machine and cleans up the interface.

https://github.com/near/core-contracts/commit/5d123106d5187bd773bc0ff19580a0f0eb946a56 is pretty optional, and it duplicates the contract logic but saves about a thousand bytes off the raw and unoptimized wasm blob. In the future, after the next SDK release, this workaround won't be needed for functions that are used in wasm and native testing targets.

cc @ilblackdragon feel free to take any of this and close this PR or merge it into yours if you've tested. I ran the tests and they still pass but I don't know how extensive they are.

Edit: also updated to SDK version 4.0.0-pre.1 to remove workaround which involves duplicate code from here [`5d12310` (#149)](https://github.com/near/core-contracts/pull/149/commits/5d123106d5187bd773bc0ff19580a0f0eb946a56)


generic-factory/Cargo.toml
```diff
@@ -18,7 +18,7 @@ panic = "abort"
 overflow-checks = true
 
 [dependencies]
-near-sdk = "3.1.0"
+near-sdk = { version = "4.0.0-pre.1", features = ["unstable"] }
```
mikedotexe, line 21: possibly update to 4.0.0-pre.2, thx for patience

mikedotexe (approved):

ChaoticTempest (approved):
LGTM, but probably want to up the version still if there's isn't any issues with that

Issue #139, opened 2021-04-15
Lockups: confusing doc about termination information


telezhnaya:
We have `TerminationInformation::unvested_amount`:
```
The amount of tokens that are unvested and has to be transferred back to NEAR Foundation.
```
https://github.com/near/core-contracts/blob/master/lockup/src/types.rs#L175

We also have `LockupInformation::termination_withdrawn_tokens`
```
The amount of tokens that were withdrawn by NEAR foundation due to early termination of vesting.
```
https://github.com/near/core-contracts/blob/master/lockup/src/types.rs#L36

In the real example, their sum is equal to `LockupInformation::lockup_amount`. 
Are we sure that both these amounts will go to foundation?
Looks like `TerminationInformation::unvested_amount` is actually an amount that could be taken by user, because it was vested before termination.

Issue #146, opened 2021-05-19
Lockup: mark some fields as deprecated, fix comments


telezhnaya:
While I refactored the doc, I found out that there are many deprecated things such as:
1. [`lockup_duration`](https://github.com/near/core-contracts/blob/master/lockup/src/lib.rs#L185)
2. [`transfers_information`](https://github.com/near/core-contracts/blob/master/lockup/src/lib.rs#L187)

It's a good idea to mark it explicitly in the code not to confuse the users.

Some things should be also renamed:
1. `foundation_accout_id` could be `terminator_account_id` (took it from [this discussion](https://github.com/near/core-contracts/pull/144/files#r634532162))

Some comments are confusing or even wrong:
1. about [`lockup_duration`](https://github.com/near/core-contracts/blob/master/lockup/src/types.rs#L37) and [`release_duration`](https://github.com/near/core-contracts/blob/master/lockup/src/types.rs#L40) (tokens unlocking moment is not always the moment of enabling transfers)
2. about [Transfers Information](https://github.com/near/core-contracts/blob/master/lockup/src/types.rs#L53)
3. To be continued. At least, we should re-check the whole [types.rs](https://github.com/near/core-contracts/blob/master/lockup/src/types.rs) file

Since it's not a good idea to touch the code on each minor fix of comments/adding [deprecated] annotation, I think it's better to collect the issues and fix them in bulk mode.

telezhnaya:
[This issue](https://github.com/near/core-contracts/issues/139) could be also solved by improving comments here

telezhnaya:
@ailisp checked the contract, here is the result: https://hackmd.io/@ailisp/rJbiDYr5_

Issue #145, opened 2021-05-12
Reduce minimum required balance in Lockup


telezhnaya:
https://github.com/near/core-contracts/tree/master/lockup#early-vesting-termination
The last 2 paragraphs there say:

> The amount withdrawn in the event of termination by Contract Creator may be lower than the initial contract amount. It's because the contract has to maintain the minimum required balance to cover storage of the contract code and contract state. The amount of NEAR tokens locked to maintain the minimum storage balance is 35 NEAR. 35 NEAR is enough to cover storage for 350000 bytes on Mainnet at the price of 1 NEAR per 10000 bytes.
> 
> If there is still a termination balance deficit due to the minimum required balance, the owner may decide to fund the deficit on this account to finish the termination process. This can be done through a regular transfer action from an account with a liquid balance.
> 

We need to:
- [ ] add to Lockup Factory the step with the transfer to lockups additional NEAR to cover these operations (TBD: all lockups? at the moment of creation? at the moment of termination?)
- [ ] storage prices become 10 times smaller, we should change 35 NEAR to 3.5 NEAR

telezhnaya:
@AngelBlock @yessins let's decide whether we want to implement a second task here

PR #153, opened 2021-07-02
Fix lockup issues found by @ailisp

telezhnaya:
The first commit solves #145 
The second commit adds the instruction on how to build the contract.

All the other commits are addressing issues in [the report](https://hackmd.io/@ailisp/rJbiDYr5_) file-by-file, you could check them separately if you want to.

I also partially fixed issues from https://github.com/near/core-contracts/issues/146, some of them were intersected with the issues from the report.

---

What **haven't** been done during this PR:
- I didn't fix the comments in README. You could see some fixes there because they are required to have a consistent state with the code, but I will re-check the comments from @ailisp separately
- I didn't fix https://github.com/near/core-contracts/issues/139 because I don't feel confident in it. @ailisp does not find the problem there at all. I feel I need more time to dig into that, not sure the issue worth it.
- I didn't mark as deprecated `transfers_information`, `voting`, and other related stuff. While I was rewriting README on the previous iteration, @evgenykuzyakov pointed out that it's not correct to say that transfers were enabled on Oct 13, 2020, because the contract could be invoked not only in the Mainnet, and that date could differ. So, from that point of view, the mechanism should not also be deprecated, it still could be used.
- I didn't rename `foundation_account_id` to `terminator_account_id` because it changes the public interface, so it could break the backward compatibility.

----
@AngelBlock you could be also interested in it


lockup/src/lib.rs
```diff
@@ -315,8 +313,8 @@ mod tests {
                     vesting_schedule,
                     salt: SALT.to_vec().into(),
                 }
-                .hash()
-                .into(),
+                    .hash()
```
ailisp, line 316: I'm not sure, is this formatting right?

ailisp (approved):
Lgtm! Nice job

lockup/src/lib.rs
telezhnaya: CLion fixes it that way, and I've seen other places with the same formatting, so I decided to have these changes

telezhnaya:
Rebased because we merged required PR (nothing is changed in this PR)
Now diff is exactly as it should be

evgenykuzyakov (approved):

Issue #154, opened 2021-07-02
Unable to build the contracts with the provided Dockerfile


telezhnaya:
It should be possible to build and test all the contracts with the Dockerfile from `near-sdk-rs`. 
Actually, at least with any Dockerfile, but let's dream we have one env for all.

For the build, I use `./build.sh`. For the test, I use `cargo test`. If I say "works fine", I mean "both build and test finished successfully".

We already fixed the Dockerfile so that `lockup` contract could be built on top of that: https://github.com/near/near-sdk-rs/pull/454
The good news is that we accidentally fixed `w-near` with this Dockerfile.

For these contracts, `./build.sh` fails because toolchain is not OK, `cargo test` fails because it requires nightly, both with the old and the new version of Dockerfile:
- `exchange-deposit-receiver`
-  `multisig`
- `staking-pool`
- `staking-pool-factory`

For these contracts, `./build.sh` fails because toolchain is not OK, `cargo test` is OK, both with the old and the new version of Dockerfile:
- `lockup-factory` 
- `voting`
- `whitelist`

It's important to fix the contracts or the environment.
The root README should be also rewritten after that.

telezhnaya:
another TODO here: add the instruction how to ann new contract
need to add the contract name to `CONTRACTS` file (see #159)

PR #151, opened 2021-06-22
[Lockup]: Upgrade to near-sdk version 3.1.0

evgenykuzyakov:
- Updating `near-sdk` version to `3.1.0`
- Reimplementing simulation tests using `near-sdk-sim`
- Removing `rust-toolchain` as well toolchain from `build.sh`, because contract-builder from `near-sdk-rs` uses fixed Rust version, instead of the toolchain.
- Recompiling lockup with contract builder from `near-sdk-rs` using Docker.

## Test plan:
- Unit tests + simulation tests

PR #156, opened 2021-07-03
Rebuild contracts

telezhnaya:
As we deleted toolchain in #151, we need to update `./build.sh` scripts and wasm binaries.

I decided to fix at least the obvious part. 

With these changes, these contracts could be successfully built and tested via `cargo test`:
- `lockup-factory`
- `voting`
- `whitelist`
- (and `lockup` was fixed in #151)

These contracts still require to fix `cargo test`, but building work fine:
- `multisig`
- `staking-pool`
- `staking-pool-factory`

`exchange-deposit-receiver` requires to fix both build and test, since it wants to use nightly.

I partially fixed https://github.com/near/core-contracts/issues/154 here

We also need to fix README (there are words about the toolchain), but since we haven't fixed all the contracts, I have nothing to say there for now.

@evgenykuzyakov could you please help to fix the build of `exchange-deposit-receiver` and the tests in `multisig`, `staking-pool`, `staking-pool-factory`, `exchange-deposit-receiver`? Ideally, we want to get rid of nightly.


bowenwang1996 (approved):

evgenykuzyakov:
Building factories contracts require building the template contracts first. Please double check that you've built them in the correct order. Otherwise looks good.

As far is `exchange-deposit-receiver`, we can delete it completely, since it's was deprecated in favor of implicit accounts IDs.

ilblackdragon:
Let's make sure contracts are re-built inside Docker container with neutral path, and that container build files are included in the repo (right now they are not at least for some of the contracts). Example of such build file (with cache) can be found here - https://github.com/ref-finance/ref-contracts/blob/main/ref-exchange/build_docker.sh

telezhnaya:
@ilblackdragon will do that in the separate issue https://github.com/near/core-contracts/issues/164

Issue #155, opened 2021-07-02
Set up CI for the contracts


telezhnaya:
It would be great to have a check that the code should pass `cargo test` successfully before merging it to master.
Another check that I really want to add is that `./build.sh` for each contract produces the same wasm file as the file that was committed last time (a check that the developer did not forget to update it).

Most of the contracts have issues with building/testing them, I checked them all, the result is here https://github.com/near/core-contracts/issues/154

But anyway, we could set up the CI in parallel.
@chefsale do you need any help/scripts for doing that? I see we have some https://github.com/near/core-contracts/tree/master/scripts
But I haven't tried them yet

telezhnaya:
@chefsale thank you again! This is awesome

PR #159, opened 2021-07-05
feat(buildkite): added initial core-contracts CI setup

chefsale:
Creates a pipeline in buildkite which makes sure that we:

1) Run tests on every contract specified in the CONTRACTS file 
2) Check in all the wasm files on code changes specified in the CONTRACTS file 

based on the issue: https://github.com/near/core-contracts/issues/155

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,10 @@
+steps:
+  - label: "check artifacts"
+    command: |
+      source ~/.cargo/env
+      ./scripts/build_all.sh
+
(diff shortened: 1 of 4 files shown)


telezhnaya:
Thank you @chefsale! I will look at it deeper after we resolve the existing issues with building and testing

telezhnaya (approved):
I am not sure I'm the right person to check it, but looks good to me 🙂

chefsale:
@bowenwang1996 Can you please review this PR?

PR #161, opened 2021-07-07
Remove exchange-deposit-receiver contract

telezhnaya:
@evgenykuzyakov mentioned that `it's deprecated in favor of implicit accounts IDs`
https://github.com/near/core-contracts/pull/156#issuecomment-874243142

@bowenwang1996 @nearmax @evgenykuzyakov please approve. If you know someone else who can disagree with this decision, please add them to the reviewers.

Build and test are broken (we need to configure the environment and fix library versions).
If we don't need it anymore, it's easier to remove it completely.


ilblackdragon (approved):

ilblackdragon:
Indeed, it was decided to not go this route and use implicit accounts instead.

PR #163, opened 2021-07-12
feat(core-contracts): Add build scripts for Docker

telezhnaya:
@ilblackdragon suggested adding the scripts for Docker https://github.com/near/core-contracts/pull/156#issuecomment-877596236

For now, it does not work because of https://github.com/near/near-sdk-rs/issues/465

Comment for myself
After the container on DockerHub will be updated, do not forget to check that:
- [ ] the files are executable after cloning;
- [ ] all the contracts are built successfully;
- [ ] I haven't fucked up with the final place of wasm binary

### DIFF
--- a/lockup-factory/build_docker.sh
+++ b/lockup-factory/build_docker.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# Exit script as soon as a command fails.
+set -e
+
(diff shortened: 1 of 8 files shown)


chefsale:
I believe we should just use one script to build these instead of copy pasting it around, all the contract specific things should still be configurable via variables.

chefsale:
I believe we can close this one now?

PR #167, opened 2021-07-12
feat(lockup): bump the version to 3.1.0, fix meta docs

telezhnaya:
We will merge this after the CI will be set up correctly


telezhnaya:
Maybe we need to add a warning to README-s of `multisig`, `staking-pool`, `staking-pool-factory`? Something like
```markdown
*The CI is switched off for this contract, fixing tests is required*
```

frol (approved):

PR #166, opened 2021-07-12
feat(buildkite): added a build_all_docker script and --check flag for the artifacts in the CI pipeline

chefsale:

PR #168, opened 2021-07-13
Add build script for docker

telezhnaya:
Instead of #163, we can use general solution

@chefsale you can invoke this script in a loop, passing there contract names from `CONTRACTS` file

### DIFF
--- a/scripts/build_docker.sh
+++ b/scripts/build_docker.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Exit script as soon as a command fails.
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# Pass contract name as an argument
+NAME="$@"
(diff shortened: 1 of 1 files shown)


ilblackdragon:
Are we planning to include built wasm files in the repo here? Or the built versions will be published as artifacts of the CI pipeline?

telezhnaya:
@ilblackdragon All the binaries will be in the same place where they are now.
It's done by this line https://github.com/near/core-contracts/pull/168/files#diff-b7237a9b3044006158e6f45c546f2f128efd7cef9c4598d4c91be572218cc58cR33

chefsale:
I think eventually we can move it to be published as artifacts, but first goal is to make sure that cirrent binaries are not stale and properly updated. Once we have that we can look into publishing the on github releases or some artifactory.

chefsale (approved):
LGTM, let's land this and than I can update my PR: https://github.com/near/core-contracts/pull/166 to make the `--check` flag pass using this docker build instead.

PR #171, opened 2021-07-16
State cleanup contract

ilblackdragon:
Allows to cleanup state of an account without deleting it.
Deploy this contract instead of the current one, clean up state and can redeploy new contract or use account in other ways.


state-cleanup/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+struct Contract {}
+
+#[derive(Serialize, Deserialize)]
+#[serde(crate = "near_sdk::serde")]
+struct Args {
+    pub keys: Vec<String>,
+}
+
+#[no_mangle]
+pub extern "C" fn clean() {
```
bowenwang1996, line 17: There is no guarantee that this can finish within gas limit?

frol (approved):

state-cleanup/src/lib.rs
ilyar: I think there is no guarantee here, it is necessary to first calculate the total size of the deleted data before sending the array of keys

state-cleanup/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+    env::setup_panic_hook();
+    env::set_blockchain_interface(Box::new(near_blockchain::NearBlockchain {}));
+
+    let input = env::input().unwrap();
+    let args: Args = serde_json::from_slice(&input).unwrap();
+    for key in args.keys.iter() {
+        env::storage_remove(&base64::decode(key).unwrap());
```
austinabell, line 24: Optional, but can use [the `Base64VecU8` type](https://docs.rs/near-sdk/3.1.0/near_sdk/json_types/struct.Base64VecU8.html) instead of `String` to just handle decoding this during deserialization

state-cleanup/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+use near_sdk::base64;
+use near_sdk::serde::{Deserialize, Serialize};
+use near_sdk::serde_json;
+use near_sdk::{env, near_bindgen};
+
+#[allow(dead_code)]
+#[near_bindgen]
+struct Contract {}
```
austinabell, line 8: Should just be able to delete this? Or am I forgetting about some dependency from v3 or otherwise?

austinabell (commented):
This might be a good candidate to be a no_std contract. I'll port it over as an alternative when I have free time

state-cleanup/src/lib.rs
ilyar: agree i ran this code on version `4.0.0-pre.2`
```rust
use near_sdk::serde::Deserialize;
use near_sdk::serde_json;
use near_sdk::env;
use near_sdk::json_types::Base64VecU8;

#[derive(Deserialize)]
#[serde(crate = "near_sdk::serde")]
struct Args {
    pub keys: Vec<Base64VecU8>,
}

#[no_mangle]
pub extern "C" fn clean() {
    env::setup_panic_hook();
    let input = env::input().unwrap();
    let args: Args = serde_json::from_slice(&input).unwrap();
    for key in args.keys.iter() {
        env::storage_remove(&key.0);
    }
}
```

ilyar: and `3.1.0`

frol: FYI, I have no idea why that might be necessary

ilyar: it helped me to solve the error: 

> The collection is an inconsistent state. Did previous smart contract execution terminate unexpectedly?

to be more precise it helps to better understand how storage works.
I left a note for myself, it may be useful not only for me https://gist.github.com/ilyar/19bdc04d1aa09ae0fc84eb4297df1a1d#file-clear_state_of_contract_on_near_protocol-md

MaximusHaximus:
Howdy folks,
Is there anything pending blocking this contract from being merged and becoming an official core contract?  We would like to use it to ensure that when we disable a multisig contract used for 2FA by the NEAR wallet, we first clear any on-chain state that the contract is maintaining.  Ref: near/near-api-js#736

MaximusHaximus:
Howdy! Just checking in -- is there any reason why this contract can't be used at this point? Can we get this PR merged?

esaminu:
Hello all, we have another report of a user being affected by uncleared state after deleting the multisig contract. Can this be used in the wallet as per @MaximusHaximus's suggestion?

agileurbanite:
@austinabell @ilyar @bowenwang1996 @frol any updates on whether the the status of this PR?  @frol has already approved but seeing feedback from @austinabell wondering if these are nits that can be addressed in a separate PR or if these are blockers.  Thanks!

austinabell:
> but seeing feedback from @austinabell wondering if these are nits that can be addressed in a separate PR or if these are blockers.  Thanks!

No blockers! Those are nits

jberrytech:
If we're feeling good about this, @austinabell, could you execute the merge?

austinabell (approved):

austinabell:
Opened #203 for how to optimize this in the future

deveusss:
How to get all the keys for contract with large state, it gives an error 
Server error: State of contract contract.near is too large to be viewed

wackazong:
For other people landing here: you can prevent the error State of contract _____.near is too large to be viewed  when calling view-state via the JSON RPC API if you select an alternative RPC provider from this list (also works for testnet if you replace testnet with mainnet in some of the URLs): [docs.near.org/api/rpc/providers](https://docs.near.org/api/rpc/providers) I was successful with OOMNIA: [endpoints.omniatech.io/v1/near/testnet/public](https://endpoints.omniatech.io/v1/near/testnet/public)

PR #172, opened 2021-07-26
chore: update lockup contract to 4.0 pre-release

austinabell:
there shouldn't be any other major changes before the stable release but might want to wait to move this in until 4.0 is released just to be sure.


austinabell:
I can't see why the build fails and I can't repro any failure locally with testing or formatting

PR #176, opened 2021-07-27
chore: update staking pool factory to 4.0.0-pre.1

austinabell:
ditto #172


ilblackdragon:
Does this pass a clean build without lock file? Because I think we removed some of the dependencies that are in tests like near-runtime-standalone from nearcore?

PR #188, opened 2021-11-11
fix(lockup): add test of unlocking schedule with termination

### LINKED ISSUE #187: Add test in lockup contract about termination process
1. Suppose we have two lockup contracts A and B that had exactly the same initialization parameters. These contracts are also interacted in exactly the same way and moment by the users, i.e. they receive the same transactions/receipts at the same moments;
2. Suppose B is getting terminated at a certain timestamp T;
3. Is the following true for any possible situation: contracts A and B will unlock exactly the same number of tokens (that also become immediately liquid/withdraw-able) at exactly the same moments from timestamp T' (T'<=T) until some timestamp T'' (T<=T''), and after timestamp T'' B will not unlock any tokens.

Me and @frol came to the conclusion that it's true, but it's better to add a test for that

telezhnaya:
Closes #187 

I also need to rebuild the binary again; I'll do that after we review the test.

I decided to check all that we wanted on one contract instead of 2, it looks simpler for me and it checks the same stuff.

### DIFF
--- a/lockup/src/lib.rs
+++ b/lockup/src/lib.rs
@@ -1726,6 +1726,163 @@ mod tests {
     }
 
     #[test]
+    fn test_unlocking_schedule_with_termination() {
+        // This test checks that termination of vesting
+        // fixes the amount of tokens that finally should become liquid (*),
+        // but does not change the schedule of unlocking the tokens.
+        // The contract should act as there was no termination at all,
+        // until we do not reach the amount (*).
+        // After we reach this amount, the unlocking process should stop.
+
+        let mut context = basic_context();
+        testing_env!(context.clone());
+
+        let lockup_amount = to_yocto(1000);
+        let vesting_schedule = new_vesting_schedule(100);
+        let mut contract = new_contract(
+            true,
+            Some(vesting_schedule.clone()),
+            Some(to_nanos(YEAR * 3).into()),
+            true,
+        );
+
+        // Everything is locked & unvested in the beginning
+        assert_eq!(
+            contract.get_vesting_information(),
+            VestingInformation::VestingHash(
+                VestingScheduleWithSalt {
+                    vesting_schedule: vesting_schedule.clone(),
+                    salt: SALT.to_vec().into(),
+                }
+                    .hash()
+                    .into()
+            )
+        );
+        assert_eq!(contract.get_owners_balance().0, 0);
+        assert_eq!(contract.get_liquid_owners_balance().0, 0);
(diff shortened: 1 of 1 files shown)


telezhnaya:
I guess it's better to mention the doc in the comments
https://wiki.near.org/getting-started/near-token/lockups#termination-of-vesting
It could be useful for the reader to understand the logic better

lockup/src/lib.rs
```diff
@@ -1725,6 +1725,163 @@ mod tests {
...
+                status: TerminationStatus::ReadyToWithdraw,
+            })
+        );
+
+        // Moving further, to the moment when unlocking has already started
+        context = get_context(
+            system_account(),
+            to_yocto(LOCKUP_NEAR),
+            0,
+            to_ts(GENESIS_TIME_IN_DAYS + 400),
```
frol, line 1837: I find this 400 to be an odd date. It is better to choose checkpoints that are easy to reason about:

* the day of vesting termination
* the next day after it
* the day before the expected day of all non-terminated tokens to be unlocked
* the last day of unlock
* the day after the day of unlock

lockup/src/lib.rs
```diff
@@ -1725,6 +1725,163 @@ mod tests {
...
+            locked_tomorrow < locked_today,
+            "Locked amount should decrease"
+        );
+
+        // Moving further, to the moment when unlocking should stop
+        context = get_context(
+            system_account(),
+            to_yocto(LOCKUP_NEAR),
+            0,
+            to_ts(GENESIS_TIME_IN_DAYS + 720),
```
frol, line 1865: Why 720? I feel it should stop unlocking the tokens vested over `YEAR - vesting_cliff_offset_in_days + termination_offset_in_days` (YEAR - 100 + 200 = 465 days of vested tokens out of 4 years vesting contract), and that should happen on `GENESIS_TIME_IN_DAYS + lockup_duration + 465`th day (given that lockup_duration is 365, I expect it to be +830) in lockup contract v3 implementation (without lockup cliff), and on `GENESIS_TIME_IN_DAYS + 465`th day in lockup contract v2 implementation (where lockup had cliff).

What do I miss?

frol (commented):
@telezhnaya I find this test quite hard to follow due to "magical consts"

ilblackdragon:
Standard wise, it will be test(lockup) if there is no actual fix in the PR but only tests added.

telezhnaya:
Addressed all the review comments in https://github.com/near/core-contracts/pull/191

Issue #184, opened 2021-10-20
Sweep through multisig2 contract for efficiencies


mikedotexe:
There is a pull request in draft at the time of this writing: #183 
This is a modification of the extended multisig located on the branch: 
https://github.com/near/core-contracts/tree/extend-multisig

The task at hand in this issue is to go through and follow all the best practices located under the Small Wins section of the SDK Docs here:
https://www.near-sdk.io/reducing-contract-size/examples#small-wins

Namely, anywhere where there's an `assert!`, `assert_eq!`, or `assert_ne!` we must replace them as described in that doc.
Similarly, every time there's an `expect` in the code, replace is as described there as well.

This will reduce the size of the contract, which is very important to be as small as possible.
Please open up a pull request against the branch `extend-multisig-update-sdk`

PR #189, opened 2021-11-11
reduce multisig2 contract size by replacing panic code with env::panic_str(...)

adsick:
Addresses #184
the gain in size is not huge but the changes made multisig2.wasm 735 bytes lighter.


multisig2/src/lib.rs
```diff
@@ -449,17 +457,17 @@ mod tests {
...
-    use near_sdk::Balance;
 
     use super::*;
 
     /// Used for asserts_eq.
     /// TODO: replace with derive when https://github.com/near/near-sdk-rs/issues/165
     impl Debug for MultiSigRequest {
         fn fmt(&self, _f: &mut Formatter<'_>) -> Result<(), Error> {
-            panic!("Should not trigger");
+            env::panic_str("Should not trigger");
```
ilblackdragon, line 470: you don't need to update tests, as they are not contributing to the size of the contract.

multisig2/src/lib.rs
```diff
@@ -127,11 +128,12 @@ impl MultiSigContract {
     /// @params num_confirmations: k of n signatures required to perform operations.
     #[init]
     pub fn new(members: Vec<MultisigMember>, num_confirmations: u32) -> Self {
-        assert!(!env::state_exists(), "Already initialized");
-        assert!(
```
ilblackdragon, line 131: it's easier to replace this with to keep the same number of instructions but keep the code pretty much the same.
```
#[inline]
fn assert(condition: bool, error: &str) {
    if condition {
        env::panic_str(error);
    }
}
```

ilblackdragon (commented):
Will merge this in but leaving comments for future.

multisig2/src/lib.rs
adsick: ah, yea, was thinking about that too but decided to go all in)

ilblackdragon:
It's merged directly into #125

PR #191, opened 2021-12-09
test(lockup): add test of unlocking schedule with termination

telezhnaya:
Warn: it should never be merged, it's based on v2 of lockup contract

The reason of this code is to test the lockup functionality that is used most widely nowadays. We want to be sure how current lockups work

### DIFF
--- a/lockup/src/lib.rs
+++ b/lockup/src/lib.rs
@@ -1778,25 +1778,25 @@ mod tests {
         let vesting_nanos_total =
             (vesting_schedule.end_timestamp.0 - vesting_schedule.start_timestamp.0) as u128;
 
-        let expected_vested_amount_at_cliff_day =
-            lockup_amount / vesting_nanos_total * vesting_nanos_passed;
+        let expected_vested_amount_at_cliff_day = (U256::from(lockup_amount)
+            * U256::from(vesting_nanos_passed)
+            / U256::from(vesting_nanos_total))
+        .as_u128();
         let vested_amount_at_cliff_day = contract
             .get_locked_vested_amount(vesting_schedule.clone())
             .0;
-        assert_almost_eq_with_max_delta(
+        assert_eq!(
             expected_vested_amount_at_cliff_day,
-            vested_amount_at_cliff_day,
-            to_yocto(1),
+            vested_amount_at_cliff_day
         );
         assert_eq!(to_yocto(250000), vested_amount_at_cliff_day);
 
         let expected_unvested_amount_at_cliff_day =
             lockup_amount - expected_vested_amount_at_cliff_day;
         let unvested_amount_at_cliff_day = contract.get_unvested_amount(vesting_schedule.clone()).0;
-        assert_almost_eq_with_max_delta(
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


frol:
> Warn: it should never be merged, it's based on v2 of lockup contract

Hmm, how is it based on v2 if the code is from the master branch?

I suggest we create lockup-v2 branch that corresponds to `lockup-v2.0.0` tag, and we create a PR with `lockup-v2-extra-tests` branch, and once we review it, we can close the PR (while keeping the branch), and also update the README on master with the link to all those tags and branches to clarify the situation there.

Also, we need the same test on master (lockup v3), so we can at least demonstrate how the behavior is different between the versions.

lockup/src/lib.rs
```diff
@@ -1662,6 +1662,317 @@ mod tests {
...
+
+        // *** day 1066, 1 day after unlock stopped ***
+        let ts_1_day_after_unlock_stopped = to_ts(GENESIS_TIME_IN_DAYS + YEAR + 201);
+        context.block_timestamp = ts_1_day_after_unlock_stopped;
+        testing_env!(context.clone());
+
+        assert_eq!(
+            contract.get_locked_amount().0,
+            unvested_amount_at_termination_day
+        );
```
frol, line 1973: Please, add assert on `get_unlocked() == initial lockup amount - terminated`

telezhnaya: Sorry, I will not add this check because both possible methods are broken
https://github.com/near/core-contracts/issues/192#issuecomment-996766606
I guess they will work for this specific case, but anyway, I don't want to use them, that's a matter of coincidence that they give us the expected result here

P.S. It's still too early to review this; I will ping you separately when I finish the work here

telezhnaya:
`test_unlocking_schedule_with_termination_and_vesting_started_before_phase2`
Picture for test 1

telezhnaya:
`test_unlocking_schedule_with_termination_and_vesting_started_after_phase2`
Test 2

telezhnaya:
We discussed some ideas on how to improve these tests. One of the ideas was to make the same tests on the master branch.
I'm not sure it makes sense with the same setup. `lockup_duration` is deprecated. Moreover, not sure someone will lock the tokens before vesting starts.
I think it makes sense to write one test on master, where lockup is not connected to Phase II AND lockup goes after vesting. @frol what do you think?

This PR is ready to review.

telezhnaya:
Actually, I'm not sure whether we need these tests on master at all. I hope no one will use this contract right now; it means that we can stop spending time on it and invest it in other issues. 
It was important to make the work of v2.0.0 more transparent, and I hope these tests help with it at least somehow.

lockup/src/lib.rs
frol: As per our last discussion, let's add this assert anyway, and if it crashes (supposedly in the next test), let's comment it out and leave an explanation why we believe it is not critical

telezhnaya:
Created the test for v3.0.0
https://github.com/near/core-contracts/commit/9b08fe4c6e90b0effffe831363b9673102ba6c27

PR #204, opened 2022-03-02
ci(check): add missing w_near wasm binary

chefsale:


rtsai123 (approved):
Thanks for adding the missing binary

austinabell (approved):
confused, the binary was there, was it an outdated one or does this revert it back to the one that's deployed?

chefsale:
There is a check in the CI which makes sure that the wasm binary is the actual commited code. Somebody who made the change didn't update the binary.

PR #205, opened 2022-03-03
bumping rust version to 1.59.0 to fix compiling issues requiring Rust 2021

rtsai123:
Bumping rust version to 1.59.0 stable to resolve the following github issues:
[Github Issue 782]( https://github.com/near/near-ops/issues/782)
[Github Issue 730]( https://github.com/near/near-ops/issues/730 )
[Github Issue 727]( https://github.com/near/near-ops/issues/727)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,13 +2,13 @@ steps:
   - label: "check artifacts"
     command: |
       source ~/.cargo/env
-      rustup default 1.51.0
+      rustup default 1.56.0
       rustup target add wasm32-unknown-unknown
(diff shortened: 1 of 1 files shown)


austinabell (approved):

anshal-savla (approved):
I have left a question in the comments.

rtsai123:
> 

I don't think we do save older builds from what I can tell.

austinabell:
> i am just seeing one image in the docker repo tagged latest. Do we not save older builds?

No, but I don't think there is a reason we shouldn't improve this to make it more usable. Currently the docker image isn't used very much externally it seems

PR #208, opened 2022-03-31
feat: implement state replace contract

### LINKED ISSUE #203: Optimize state clearing contract
Based on some comments made here: https://github.com/near/core-contracts/pull/171#pullrequestreview-767219536

There are a few things that can be done to make this smaller (without changing the interface at all):
1. Update to 4.0 pre-release of SDK to avoid having to go through `blockchain_interface` for syscalls
2. Replace serde/serde-json with something like https://github.com/dtolnay/miniserde. This will make the contract to deploy much smaller, but deserializing might use slightly more compute/gas. This change is likely worth it in the vast majority of cases.
3. Make it `no_std`
4. Replace unwraps with manually handled errors through syscalls to avoid putting file-specific data into wasm binary

austinabell:
closes #203 

cc @esaminu does this fit the criteria for what wallet needs?

What has changed:
- interface changed from `'{"keys":["...", "..."]}'` to `'["...", "..."]'` to avoid redundant strings.
  - This won't work with near-cli because of this https://github.com/near/near-cli/issues/920 so I'm happy to add the redundant data
- Changed to no_std contract -- size down from 96k -> 67k (70k with new functionality)
  - Extra boilerplate, would be nice to use https://github.com/austinabell/nesdie to avoid, but wanted to keep minimal dependencies for security
  - Can switch to using SDK when a middle layer for no_std is implemented
- Optimized deserialization -- ~~no longer allocates or copies values in any way~~ Edit: it does now because the stream deserialization was not correct, I'll look into it later
  - Previously allocated every single encoded key string as well as a `Vec` buffer of all strings, now none of that -- optimized for gas usage

Did not change the serialization from serde because it was indicated JSON is preferred. Can switch to using https://github.com/dtolnay/miniserde for very small code size, but it will come at the cost of less efficient deserializations. If gas optimizations are important, can open an issue to swap serialization by a feature and compare the differences.


esaminu:
Looks good! We'd have to update [how we call it](https://github.com/near/near-api-js/pull/802/files#diff-823d20b61d17e5931208a2a8e2ba6c7088e41396df9794127b6b91ff2e33207bR355) in `near-api-js` due to interface changes but aside from that we should be good to go. For the state restoration operation suggested [here](https://nearinc.atlassian.net/wiki/spaces/WA/pages/15302661/2FA+invalid+state+issue+resolution#Remediation-for-highly-valuable-affected-accounts%3A) we can just use one transaction with function calls to first `clean` and then `storage_write` to add the new state in near/near-wallet#2561.

cc: @MaximusHaximus

austinabell:
hmm, hold up on merging this though, adding workspaces tests seemed to brick the build -- diving into why that affected it

esaminu:
Awesome! Can this be reviewed and merged? @MaximusHaximus @agileurbanite

esaminu:
@austinabell after further discussion with the team, we think it would be better to leave the `clean` method signature as it currently is due to serialization issues with array args in `near-api-js`. Aside from we'd love to see this merged!

austinabell:
> @austinabell after further discussion with the team, we think it would be better to leave the `clean` method signature as it currently is due to serialization issues with array args in `near-api-js`. Aside from that we'd love to see this merged!

Done :) Sorry for being opinionated and trying to avoid tech debt for this.

format is now {"keys":...} for clean and {"entries: [[.., ..], ...]} for replace, as readme shows.

Also updated workspaces and the tests now that that's released so this is good timing :D

esaminu:
Awesome! Would love to see this merged :)

austinabell:
Is there anything blocking this? Who is reviewing/approving this change?

frol (approved):

PR #210, opened 2022-04-12
fix(multisig2): access keys not able to call add_request_and_confirm

camchis:
access keys were being given access to call `add_and_confirm_request` which doesn't exist, the correct function name is `add_request_and_confirm`


miraclx (approved):

miraclx (changes requested):
Can you update the wasm file?

camchis:
> Can you update the wasm file?

Done! :)

PR #211, opened 2022-05-03
deploy_core.sh: fix syntax error

Mic92:


scripts/deploy_core.sh
```diff
@@ -11,7 +11,7 @@ if [ -z "${MASTER_ACCOUNT_ID}" ]; then
   exit 1
 fi
 
-if [ -z "${FOUNDATION_ACCOUNT_ID}"]; then
```
Mic92, line 14: The missing space after the string is the core issue.

PR #212, opened 2022-05-12
fix MULTISIG_METHOD_NAMES

prih:


camchis:
Noticed this too

#210

PR #220, opened 2022-12-19
lockup-factory: Add staking pool allowlist param to `create` function.

PierreLeGuen:
- Allows providing a custom staking pool whitelist per lockup.


lockup-factory/src/lib.rs
```diff
@@ -124,6 +125,12 @@ impl LockupFactory {
             foundation_account = Some(self.foundation_account_id.clone());
         };
 
+        // Defaults to the whitelist account ID given on init call.
+        let mut staking_pool_whitelist_account_id: AccountId = self.whitelist_account_id.clone();
```
ilblackdragon, line 129: to remove mut which can be a potential issue later can be just

let staking_pool_whitelist_account_id: AccountId = if let Some(account_id) = whitelist_account_id { account_id.into() } else { self.whitelist_account_id.clone() };

ilblackdragon (approved):

PierreLeGuen:
@ilblackdragon  - I was able to build the wasm file using `nearprotocol/contract-builder:latest-arm64` docker image, I've some concerns because the docker scripts are normally using the `amd-64` image. I can't see a direct impact because it's compiled to`wasm32-unknown-unknown`. Do you know any drawbacks?

(Compilation with the `amd-64` image was failing on Mac M1)

ilblackdragon:
@PierreLeGuen you didn't push that version though, right?

PierreLeGuen:
@ilblackdragon - I pushed the changes with no `mut` variable and the new wasm from `contract-builder`.

